PR #12, opened 2014-06-13
[WIP] Using curl multi select is slow for some reason.

jwage:
Trying to figure out why the Algolia PHP driver is so much slower then the JS client.

Before this change requests were taking > 1 second. After this they take 200-300 ms.


phantomevox:
Sorry for necro bumping... but, was there any conclusive update on this?

rayrutjes:
I really think that at some point this client should rely on Guzzle.

PR #29, opened 2015-01-21
Fix #26: move the version to a PHP const for performances

### LINKED ISSUE #26: Version class perform useless file_get_contents for all requests
I think all this code in **Version.php** could be replaced by a simple **const**:

```php
  public static function getValue() {
    if (!isset(self::$value)) {
      $vpath = dirname(__FILE__) . "/../../composer.json";
      if (file_exists($vpath)) {
        $composerJson = file_get_contents($vpath);
        if ($composerJson != false) {
          $composer = json_decode($composerJson);
          if ($composer != null) {
            self::$value = $composer->version;
            return self::$value;
          }
        }
      }
      self::$value = "N/A";
    } else {
      return self::$value;
    }
  }
```

Because:

- you do not need a "version" key in the composer.json file, you already use git tags and composer/packagist read them;
- this will perform disk access each time we use the client, which is bad for performances;
- a PHP constant will be better for users not using Composer;

All you have to do is to update the const when you tag a new version instead of updating the composer file.

So, new Version.php:

```php
namespace AlgoliaSearch;

class Version {
    const NUMBER = '1.5.2';
}
```

I can send the pull request if you agree with my statements.

damienalexandre:
No more filesystem access for this version header.

- also fix some minor coding std (un-used var, missing return)
- also add mbstring as composer dependency - help to detect not compatible PHP installations (mbstring is not installed by default in CentOs and all other "yum" based systems).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,6 +1,5 @@
 {
     "name": "algolia/algoliasearch-client-php",
-    "version": "1.5.3",
     "description": "Algolia Search API Client for PHP",
     "type": "library",
     "homepage": "https://github.com/algolia/algoliasearch-client-php",
@@ -20,6 +19,7 @@
         }
     ],
     "require": {
+        "ext-mbstring": "*"
     },
     "require-dev": {
         "satooshi/php-coveralls": "dev-master",
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -374,7 +374,7 @@ class Client {
                     'Content-type: application/json'
                     ), $context->headers));
         }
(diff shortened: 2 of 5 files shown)


Shine-neko:
+1

redox:
Looks good @damienalexandre, sorry for the late reply. Regarding the ```ext-mbstring``` dependency, I'm not yet sure about it -> what kind of error did you have without it?

damienalexandre:
The `mb-string` dependency is all about warning users installing the client. 

I installed the composer package localy, and it worked, but then when I got on my server, without mbstring, it failed without alerting me; composer can prevent this, just by adding this line.

redox:
:+1:  thank you @damienalexandre

PR #36, opened 2015-08-03
Use composer install to install the package

freekmurze:
This is an easier way to install that package than editing the composer.json. It'll also always pull in the latest stable version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ To setup your project, follow these steps:
 
 You can install the package via Composer:
 ```bash
-composer require You can install the package via composer:
+composer require algolia/algoliasearch-client-php
 ```
 If you don't use Composer, you can copy the `algoliasearch.php` file and the `src` and `resources` directories to your project).
 
--- a/README.md
+++ b/README.md
@@ -51,17 +51,11 @@ Setup
 -------------
 To setup your project, follow these steps:
 
-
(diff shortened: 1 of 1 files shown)


vinkla:
+1 This is definitely much better and easier to do!

freekmurze:
Fixed the composer command, sorry about that.

vinkla:
:+1:

maxiloc:
@freekmurze I not against adding the composer require way to install but I will let at least :
\---
You can also add ```algolia/algoliasearch-client-php": "1.5.* ``` to your ```composer.json```
\---
because for me it's good to have the current major version in the doc.

What do you think ?

freekmurze:
I don't think it's necessary.  New users should always require the latest stable version of the package. `composer require` does that for you.

(isn't the latest current version 1.6?)

maxiloc:
@freekmurze Yes your are right we are at 1.6
It will avoid having to maintain the version number there

redox:
@maxiloc Don't forget to apply the change in https://github.com/algolia/algoliasearch-client-readme-generator/blob/master/include/PHP/setup.snippet, otherwise we'll lose the update the next time we rebuild our READMEs.

maxiloc:
@redox thanks will do

PR #38, opened 2015-09-07
Bugfix + Optimize deleteByQuery

R2c:
Bugfix :
* Now retrieve only objectID not all object data
* Now using 1000 per package, not the default search limit (20)

Optimize :
* Optimize a little PHP
* Standard coding
* Do not wait the last call (If you have just one call, you never wait)

Sorry i can't test. You can C/C the function if you prefer and not merge..

```        
$this->client = new Client(getenv('ALGOLIA_APPLICATION_ID'), getenv('ALGOLIA_API_KEY'), null, array("cainfo" => (__DIR__ . '/../../../resources/ca-bundle.crt')));
```

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -127,7 +127,7 @@ class Index {
             throw new \Exception('No list of objectID provided');
         }
         $requests = array();
-        foreach ($objectIDs as $object) {            
+        foreach ($objectIDs as $object) {
             $req = array("indexName" => $this->indexName, "objectID" => $object);
             array_push($requests, $req);
         }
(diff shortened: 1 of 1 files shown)


vvo:
Nice.

I think we will take the bugfix but as we discussed, changing the behavior on the last call could break some other clients that are relying on the fact that the deleteByQuery should be "synchronous" (they would maybe then do a .search and see that not everything was deleted).

Can you do the change on this? Otherwise I will do it, let me know.

Thanks a lot!

R2c:
Better @vvo  ?

Nagriar:
Hi @R2c,

Thanks for you PR!

You condition to remove the latest waitTask isn't exact, in really rare cases you can have the number of hits smaller than the hitsPerPage param if you trigger the timeout of the search, in that case you won't remove all the records.

R2c:
I understand. 

However the fact of waiting is clearly a big problem in our project. This solution is a non-mandatory parameter and disable by default, therefore the person must use it knowingly. In our case we prefer to take this risk, because otherwise the performance is horrible. It may be added in the function a warning message about this setting?

We use a decorator in our project, you can take what you want in this PR =)

vvo:
@R2c We missed each other on the live chat sorry. I think both @Nagriar and me are ok with your latests improvements, let's merge!

R2c:
@vvo @Nagriar : Update with warning message on PHPDoc.

Nagriar:
:+1: excellent @R2c :boom:

PR #39, opened 2015-09-24
Added support for proxy 

remybreuils:
Added support for custom curl options like CURLOPT_PROXY
On construct you can specified an array with curl options:

```php
$options = ['curl-options' => [CURLOPT_PROXY => 'tcp://1.2.3.4:80']];
$client = new Client(
            'app-id',
            'admin_api_key',
            null,
            $options
        );
```

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -34,6 +34,7 @@ class Client {
 
     protected $context;
     protected $cainfoPath;
+    protected $curlOptions = [];
 
     /*
      * Algolia Search initialization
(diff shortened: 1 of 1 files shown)


widmogrod:
:+1: Do you know ETA? 
This feature would be very useful.

k3nn7:
Is there any chance to merge it soon? :)

src/AlgoliaSearch/Client.php
```diff
@@ -57,6 +58,8 @@ function __construct($applicationID, $apiKey, $hostsArray = null, $options = arr
         foreach ($options as $option => $value) {
             if ($option == "cainfo") {
                 $this->cainfoPath = $value;
+            } elseif ($option == "curl-options" and is_array($value)) {
```
redox: I think we need to throw an explicit error if the `curl-options` option is not an array; otherwise it will be very complex to debug.

widmogrod: Also would be nice to have `curl-options` as a constant and have some test covering this feature.

maxiloc:
Closing this in favor on #41

PR #42, opened 2015-11-19
Add .gitattributes for smaller download size

vinkla:
We don't need all these files in production.


vinkla:
Any update on this?

maxiloc:
@vinkla thanks

Issue #44, opened 2015-12-21
Support user IP and user ID setting


ivank:
Your api docs state that you have support for IP forwarding and User id tracking - https://www.algolia.com/doc/faq/analytics/will-the-analytics-still-work-if-i-perform-the-search-through-my-backend 
However this sdk does not support it. Do you plan on implementing it? 
I can create a pull request for it if you guys don't have the time.

redox:
Hi @ivank,

yes you're right, that's something we should definitely add and we would be happy to merge your PR :)

PR #45, opened 2015-12-21
Add forwared ip and user token headers

ivank:
Implements the functionality from: https://www.algolia.com/doc/faq/analytics/will-the-analytics-still-work-if-i-perform-the-search-through-my-backend
fixes https://github.com/algolia/algoliasearch-client-php/issues/44

I haven't added tests since I cant configure them to run locally...

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -107,6 +107,28 @@ class Client {
         $this->context->setRateLimit($adminAPIKey, $endUserIP, $rateLimitAPIKey);
     }
 
+    /**
+     * The aggregation of the queries to retrieve the latest query uses the IP or the user token to work efficiently.
+     * If the queries are made from your backend server, the IP will be the same for all of the queries.
+     * We're supporting the following HTTP header to forward the IP of your end-user to the engine, you just need to set it for each query.
+     *
+     * @see https://www.algolia.com/doc/faq/analytics/will-the-analytics-still-work-if-i-perform-the-search-through-my-backend
+     * @param string $ip
+     */
+    public function setForwardedFor($ip) {
+        $this->context->setForwardedFor($ip);
+    }
+
+    /**
+     * It's possible to use the following token to track users that have the same IP or to track users that use different devices
+     *
+     * @see https://www.algolia.com/doc/faq/analytics/will-the-analytics-still-work-if-i-perform-the-search-through-my-backend
+     * @param string $token
+     */
+    public function setAlgoliaUserToken($token) {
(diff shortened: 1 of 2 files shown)


src/AlgoliaSearch/Client.php
```diff
@@ -482,6 +504,7 @@ public function doRequest($context, $method, $host, $path, $params, $data, $conn
                     'X-Algolia-Application-Id: ' . $context->applicationID,
                     'X-Algolia-API-Key: ' . $context->adminAPIKey,
                     'X-Forwarded-For: ' . $context->endUserIP,
+                    'X-Algolia-UserToken: ' . $context->algoliaUserToken,
                     'X-Forwarded-API-Key: ' . $context->rateLimitAPIKey,
```
redox: Shouldn't we set those headers only if they are non-blank? I don't think it will work if we send `X-Forwarded-API-Key: <NOTHING>`

redox:
hey @ivank, sorry for the holiday season lag. Just did a comment in your PR; what do you think?

src/AlgoliaSearch/Client.php
ivank: I was wondering that myself too, but thought to better not ruffle any feathers and stick with the style you guys are using. Also I'm not sure in what environment the tests are run but I can't really run them myself, so that's hard to test.
I do think most of those headers should be sent only if they have values to minimize data-over-the-wire. Do you want me to do a refactor here?

ivank: @redox what do you think?

hkdobrev:
I was able to run the tests for this branch and they all pass :smile: 

I have considerable amount of critique to the way this package is being tested, but I will save it for another issue. Specifically for this PR the tests are structured in such a way it is not easy or at least not consistent to add unit-tests just for this change instead of full-blown integration tests.

ivank:
I second @hkdobrev in that there are a lot that can be improved in this package - both in code and tests. @redox maybe just using some automated tool to convert to PSR coding style should do wonders. For the tests you could use some PSR-7 compatible package to abstract the http requests, and then mock them altogether for a much cleaner (and faster) tests.

redox:
@maxiloc can you help?

src/AlgoliaSearch/Client.php
redox: :+1:

maxiloc:
@ivank @hkdobrev Yes improving coding style and testing is definitely on our todo list. Let's merge this first. Feel free to submit a PR for PSR compliance, we'll be happy to merge it

PR #47, opened 2016-01-06
PHP7 Removed "split" Function

colepatrickturner:
Changed split to explode() as per PHP7 spec.


maxiloc:
@colepatrickturner Nice catch, thanks for contributing

colepatrickturner:
I used Zend's PHP7 Express checker and that was the only notable incompatibility for PHP7, any other incompatibilities would be in the left-to-right expression evaluation.

PR #50, opened 2016-02-05
Adding in multipleQueries request

garethtdavies:
The documented example for the multiple queries doesn't actually make a request. It should be making a $client->multipleQueries($queries) request.


maxiloc:
Nice catch. Thanks

PR #53, opened 2016-02-13
Psr complience

ivank:
First run the commands from http://cs.sensiolabs.org/ one by one to isolate changes into separate commits, so its easier to review. Then the last few commits are by-hand modifications to clear up all the errors left from `phpcs --standard=PSR2`

As this is a quick change just to bring the codebase up to PSR-2 standards, I didn't make any logic changes, and hopefully this will be quicker to look through.
I've used also the symfony fixers as well as they are generally common-sense ones, but all of this is of course open to discussion and we could revert some of those if you don't like it.


redox:
Wow that's a bunch of commits! Thanks for contributing @ivank! I'll let @maxiloc take care of the review/merge :+1:

Maybe we could also rebase the commit before the merge?

maxiloc:
@ivank Thanks for contributing

PR #55, opened 2016-02-22
Ivank psr complience

maxiloc:


vinkla:
:-1: I think we should use [StyleCI](https://styleci.io/) instead. Then we will get much nicer error reports.

vinkla:
We shouldn't commit the PHP CS Fixer library.

maxiloc:
@vinkla This is a work in progess ! :smile:
The goal here is only to avoid merging PR which are not following the CS not to have nice errors. PR should be open with an passing CS.

jbrooksuk:
StyleCI really would be beneficial here. You really definitely don't want the `php-cs-fixer.phar` file in there at all, that's a client-side dependency. A repo should not include that.

vinkla:
@maxiloc this is what StyleCI is made for. Check out an example diff: https://styleci.io/analyses/XpK4Z8

maxiloc:
No big diff for me but if it helps, let's go for it.

vinkla:
Okay, good.

vinkla:
This is the guide to add StyleCI support:

1. Sign up at https://styleci.io/
2. Go to the repositories page https://styleci.io/account#repos
3. Enable this repository.
4. Press **Show Analysis** and then **Settings**.
5. Add the following settings:
    ```yml
    enabled:
      - unalign_double_arrow

    disabled:
      - align_double_arrow
      - phpdoc_align
    ```
6. Press **Save Configuration**.
7. Press **Analyze Now**. 

This has to be done by the owner of the repository.

.styleci.yml
```diff
@@ -0,0 +1,6 @@
+preset: PSR2
+linting: true
+finder:
+  path:
+      - "src/AlgoliaSearch"
+      - "tests/AlgoliaSearch"
```
vinkla: @maxiloc you can remove this file. Instead use the default settings.

PR #56, opened 2016-02-22
add places endpoint

maxiloc:


src/AlgoliaSearch/Client.php
```diff
@@ -46,20 +46,17 @@ class Client
      * @param applicationID the application ID you have in your admin interface
      * @param apiKey a valid API key for the service
      * @param hostsArray the list of hosts that you have received for the service
+     * @param options
```
vvo: typo?

src/AlgoliaSearch/Client.php
```diff
@@ -74,6 +71,8 @@ public function __construct($applicationID, $apiKey, $hostsArray = null, $option
                     break;
             }
         }
+
+        $this->context = new ClientContext($applicationID, $apiKey, $hostsArray);
```
vvo: why do we need to add this line in the end? I don't get it (#phpnoob)

src/AlgoliaSearch/Client.php
maxiloc: No phpstorm likes it :smile:

src/AlgoliaSearch/Client.php
vvo: OK

src/AlgoliaSearch/ClientContext.php
```diff
@@ -86,6 +76,35 @@ public function __construct($applicationID, $apiKey, $hostsArray)
...
+        return [
+            $this->applicationID.'-dsn.algolia.net',
+            $this->applicationID.'-1.algolianet.com',
+            $this->applicationID.'-2.algolianet.com',
+            $this->applicationID.'-3.algolianet.com',
+        ];
+    }
+
+    private function getDefaultWriteHosts()
+    {
```
Nagriar: @maxiloc They can create,update and delete record so you need to have special write hosts for places app:
[
                 'places.algolia.net',
                 'places-1.algolianet.com',
                 'places-2.algolianet.com',
                 'places-3.algolianet.com',
             ];

src/AlgoliaSearch/PlacesIndex.php
```diff
@@ -0,0 +1,54 @@
...
+            'POST',
+            '/1/places/query',
+            array(),
+            array('params' => $this->client->buildQuery($args)),
+            $this->context->readHostsArray,
+            $this->context->connectTimeout,
+            $this->context->searchTimeout
+        );
+    }
+}
```
Nagriar: Same than for the host, the add,delete,update api keys can be used.

src/AlgoliaSearch/ClientContext.php
vvo: @Nagriar Do you mean read and write hosts for places are the same?

Nagriar: No the write hosts use 'places.algolia.net' like for a classical application. We need to specify them because the user may want to manage the apikey from the sdk.

vvo: Ok so if you specify appID being places, hosts should be computed the right way @maxiloc by default

maxiloc: We will do api keys in a second time

vvo:
LGTM

Issue #48, opened 2016-01-15
Mock API calls


bbwebconsult:
There seem to be no way to mock Algolia's API call in the php client. What do you suggest?

rayrutjes:
Hi @bbwebconsult , could you specify your use case?

bbwebconsult:
@rayrutjes We need to test our system that uses Algolia Client without actually calling the service; for functional and unit tests.
Having a way to catch sent requests (to check their content) or fake responses from Algolia (to check that our system handles them correctly) would be very useful in our test suite.

A very good example of this is [Http Guzzle's Mock](http://docs.guzzlephp.org/en/latest/testing.html) utility.

rayrutjes:
@bbwebconsult thank you for this explanation.

For unit testing your system, you will have to test your own classes in isolation and mock your dependencies I guess. You could directly mock Algolia's classes or make another level of abstraction around it.

Regarding integration testing, the purpose being to test the whole system and Algolia's infrastructure being part of it, isn't ok to actually test against the real API? Internally Algolia does a lot of job behind the scene, so faking the results would not reflect the real behaviour of the API. 

To finish you are right about requests inspection, it is actually kind of hard to hook into the transport layer and get info about the raw requests and responses. We definitely have to think about this and we might need to consider Guzzle as an abstraction for the transport in the future.

Feel free to illustrate your needs with some actual testing code of your system, so we can challenge our actual implementation.

rayrutjes:
@bbwebconsult, I hope you found a way to properly test your code! We are closing this issue for now, feel free to open a new one at any time.

florentdestremau:
@rayrutjes I am facing the same problem at the moment, I can't afford to hit the production in all of my tests, let alone fixtures...I would like to

- disable the doctrine subscriber in test env only
- return a predefined array as hits from a search

I am trying to get this done, will keep posted. One thing that would help this would be to use an interface instead of a service call so that we could override the service with dummy requests that don't use api calls.

bboure:
@florentdestremau Are you using Codeception ?
To solve this issue I used [Stub](http://codeception.com/docs/reference/Stub) from codeception

````php
$client = Stub::construct(
                '\AlgoliaSearch\Client',
                ['fakeApplicationID', 'fakeApiKey'],
                [
                    'doRequest' => function () {
                        // return here the expected hit.
                        return [];
                    }
                ]
            );
````

PR #62, opened 2016-02-24
Refactoring, mainly docblocks and CS.

### LINKED ISSUE #32: [Improvements] Tag @return is missing in PhpDoc blocks
All functions are well documented for input parameters, thanks to the [`@param` tag](http://www.phpdoc.org/docs/latest/references/phpdoc/tags/param.html) , but the [`@return` tag](http://www.phpdoc.org/docs/latest/references/phpdoc/tags/return.html) is never used to document the return type.

This information is very useful for some IDE (like PhpStorm) that are able to warn the user about potential issues. In this case, since no return type is documented, we obtain this kind of warning messages:
```
'void' method 'browse' result used less... (Ctrl+F1) 
The value of a void function is most likely used by mistake, because such functions do not have a return value.
```

Would be nice to add these tags :)

Thanks for your work!

rayrutjes:
Started correcting some docblocks, and refactored some small pieces.

Please note that I have not a deep understanding of the behaviour of this client yet, so please tell me if some changes may break something. Hope the tests will catch problems though.

- [x] Run php-cs-fixer and add a basic .php_cs config
- [x] Add missing return tags
- [x] Enhance the docblocks 
- [x] Use more fine grained params and return types
- [x] Try to refactor some pieces of code that are repeated a lot
- [x] Better format long docblocks

Do you guys see any interest in these adjustments if they do not introduce BC issues?

Please react on the changes so I can adjust the PR.

Fixes #32

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,17 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->in(__DIR__ . DIRECTORY_SEPARATOR . 'src')
+    ->in(__DIR__ . DIRECTORY_SEPARATOR . 'tests')
+;
+
+return Symfony\CS\Config\Config::create()
+    ->setUsingCache(true)
+    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
+    ->fixers([
+        'align_double_arrow',
+        'short_array_syntax',
+        '-multiline_array_trailing_comma'
+    ])
+    ->finder($finder)
+;
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -1,4 +1,5 @@
 <?php
+
 /*
  * Copyright (c) 2013 Algolia
  * http://www.algolia.com/
@@ -23,6 +24,7 @@
  *
  *
  */
+
 namespace AlgoliaSearch;
 
 /**
@@ -36,21 +38,42 @@ class Client
     const CURLOPT = 'curloptions';
     const PLACES_ENABLED = 'placesEnabled';
 
+    /**
+     * @var ClientContext
+     */
     protected $context;
-    protected $cainfoPath;
+
+    /**
+     * @var string
+     */
+    protected $caInfoPath;
+
+    /**
+     * @var array
+     */
     protected $curlConstants;
-    protected $curlOptions = array();
+
+    /**
+     * @var array
+     */
+    protected $curlOptions = [];
+
+    /**
+     * @var bool
+     */
     protected $placesEnabled = false;
 
     /**
-     * Algolia Search initialization
+     * Algolia Search initialization.
      *
-     * @param applicationID the application ID you have in your admin interface
-     * @param apiKey a valid API key for the service
-     * @param hostsArray the list of hosts that you have received for the service
-     * @param options
+     * @param string     $applicationID the application ID you have in your admin interface
+     * @param string     $apiKey        a valid API key for the service
+     * @param array|null $hostsArray    the list of hosts that you have received for the service
+     * @param array      $options
+     *
+     * @throws \Exception
      */
-    public function __construct($applicationID, $apiKey, $hostsArray = null, $options = array())
+    public function __construct($applicationID, $apiKey, $hostsArray = null, $options = [])
     {
         if (!function_exists('curl_init')) {
             throw new \Exception('AlgoliaSearch requires the CURL PHP extension.');
@@ -59,11 +82,11 @@ class Client
             throw new \Exception('AlgoliaSearch requires the JSON PHP extension.');
         }
 
-        $this->cainfoPath = __DIR__.'/../../resources/ca-bundle.crt';
+        $this->caInfoPath = __DIR__.'/../../resources/ca-bundle.crt';
         foreach ($options as $option => $value) {
             switch ($option) {
                 case self::CAINFO:
-                    $this->cainfoPath = $value;
+                    $this->caInfoPath = $value;
                     break;
                 case self::CURLOPT:
                     $this->curlOptions = $this->checkCurlOptions($value);
@@ -73,7 +96,6 @@ class Client
                     break;
                 default:
                     throw new \Exception('Unknown option: '.$option);
-                    break;
             }
         }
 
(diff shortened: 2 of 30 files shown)


redox:
Thank you for contributing @rayrutjes! I'll let @maxiloc take over regarding the review.

The tests are failing because the API keys are (indeed) not injected, don't worry.

rayrutjes:
My pleasure @redox.

Regarding the API keys I get the fact that you do not want to pollute Agolia's servers with test traffic.
However I think giving the community feedback about the tests is nice. It is a bit frustrating to get the red flag show up!

Maybe we could isolate the actual transport layer of the client so that we can mock it when doing the tests.

AWS PHP's SDK has a nice approach illustrating what I am saying, here is an interesting snippet: https://github.com/aws/aws-sdk-php/blob/0f54c98d293afda9cff7575227da2a19851a08e5/tests/UsesServiceTrait.php

This would also address #48 by the way, and lower the required time to run the test suite from a few minutes to a few seconds.

I do realize though that this would require quite some work, and that it would probably imply bumping to a new major release.

src/AlgoliaSearch/Client.php
```diff
@@ -570,18 +629,18 @@ public static function generateSecuredApiKey($privateApiKey, $query, $userToken
...
                 }
-                $tagFilters = join(',', $tmp);
+                $tagFilters = implode(',', $tmp);
                 $queryParameters['tagFilters'] = $tagFilters;
             }
             if ($userToken != null && strlen($userToken) > 0) {
                 $queryParameters['userToken'] = $userToken;
             }
-            $urlEncodedQuery = Client::buildQuery($queryParameters);
+            $urlEncodedQuery = self::buildQuery($queryParameters);
```
maxiloc: I think static would make even more sense here

src/AlgoliaSearch/Client.php
```diff
@@ -590,7 +649,7 @@ public static function generateSecuredApiKey($privateApiKey, $query, $userToken
                 if ($userToken != null && strlen($userToken) > 0) {
                     $queryParameters['userToken'] = $userToken;
                 }
-                $urlEncodedQuery = Client::buildQuery($queryParameters);
+                $urlEncodedQuery = self::buildQuery($queryParameters);
```
maxiloc: same here

src/AlgoliaSearch/ClientContext.php
```diff
@@ -95,13 +112,16 @@ private function getDefaultReadHosts($placesEnabled)
...
+     * @return array
+     */
     private function getDefaultWriteHosts()
     {
         return [
             $this->applicationID.'.algolia.net',
             $this->applicationID.'-1.algolianet.com',
             $this->applicationID.'-2.algolianet.com',
-            $this->applicationID.'-3.algolianet.com'
+            $this->applicationID.'-3.algolianet.com',
```
maxiloc: I am not a huge fan of this standard, we should for sure have the same everywhere but this rule doesn't make a lot of sense. What do you think ?

maxiloc:
@rayrutjes Looks good, I did two comments tell me what you think.

src/AlgoliaSearch/Client.php
rayrutjes: You are right, this would eventually allow a child class to override the `buildQuery` method.

src/AlgoliaSearch/ClientContext.php
rayrutjes: No problem, it all comes down to preferences. I think it somehow eases the reordering of the items or the addition of new items.

I'll give you the last word on the subject: keep the trailing commas? or not?

maxiloc: Let's keep it, but not enforce it, as this is not part of PSR-2

.php_cs
```diff
@@ -0,0 +1,12 @@
+<?php
```
maxiloc: Do we need this file in the repo ?

rayrutjes: If we leave this in the repo, we allow everyone to just run `php-cs-fixer fix` and ensure that everyone as the same code styles.

Just a suggestion though. Could be added as a good practise to the contributing guidelines of the repo.

maxiloc: :+1: Ok good for me

src/AlgoliaSearch/ClientContext.php
rayrutjes: There you go 557392a.

rayrutjes:
Hi @maxiloc , could you take a look at what has be done so far and tell me how you feel about it?

tests/AlgoliaSearch/Tests/SettingsTest.php
```diff
@@ -41,7 +41,7 @@ public function testSettingsIndex()
 
     public function testSearchFacet()
     {
-        $res = $this->index->addObjects(array(
+        $this->index->addObjects(array(
             array("firstname" => "Robin"),
```
maxiloc: We can go for [] syntax here

.php_cs
```diff
@@ -0,0 +1,16 @@
...
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->in(__DIR__ . DIRECTORY_SEPARATOR . 'src')
+;
+
+return Symfony\CS\Config\Config::create()
+    ->setUsingCache(true)
+    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
+    ->fixers(array(
```
maxiloc: same here: []

maxiloc:
Except the [] syntax of my two comments, looks really good, after this, we can merge if your done with this PR.

rayrutjes:
I made php-cs-fixer also correct the tests folder.
I also added a param to the phpunit config to avoid a coverage related exception while running the tests in phpunit 5+.

The tests runs fine on my side. Do you have a way of running the tests on this pr before merging?

maxiloc:
@rayrutjes Yes I am going to test it right now. Can you rebase your commits in the meantime ?

maxiloc:
image

All green !

rayrutjes:
Sorry for the delay, I have rebased the pr.

maxiloc:
Ok cool ! LGTM

maxiloc:
Waiting for test to pass on https://github.com/algolia/algoliasearch-client-php/pull/65

maxiloc:
@rayrutjes Thanks for contributing. Much appreciated

PR #73, opened 2016-04-28
Fix extra headers in curl HTTP request

malengrin:
Hi,

Extra headers that are added by ClientContext::setExtraHeader() do not appear in the HTTP request.
This is because curl expect params as 
```
[
  'key: value',
  ...
]
```

and not

```
[
  'key' => 'value',
  ...
]
```


JanPetr:
Hello @malengrin, 

thank you for your PR, but we already fixed this issue in another PR. So I close this one.
If you have any other fixes or improvements in mind, don't hasitate and send PR or open an issue.

Have a nice day ;)

PR #76, opened 2016-05-05
Fix custom headers

Nagriar:
With the current code, the custom headers were always ignored because of the wrong format generated by the array_merge:

    0 => string 'X-Algolia-Application-Id: ME'
    1 => string 'X-Algolia-API-Key: TOP-SECRET'
    2 => string 'Content-type: application/json'
    'X-Forwarded-For' => string '159.253.242.85'

But in the documentation of curl it's written that only array are allowed, the associative **keyStr => val** array are not allowed.

> 	An array of HTTP header fields to set, in the format array('Content-type: text/plain', 'Content-length: 100')

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -806,7 +806,7 @@ class Client
         //curl_setopt($curlHandle, CURLOPT_VERBOSE, true);
         $curlHeaders = [];
         foreach ($context->headers as $key => $value) {
-            array_push($curlHeaders, $key . ": " . $value);
+            $curlHeaders[] = $key . ": " . $value;
         }
         if ($context->adminAPIKey == null) {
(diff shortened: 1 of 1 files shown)


src/AlgoliaSearch/Client.php
```diff
@@ -804,12 +804,16 @@ public function doRequest(
         }
 
         //curl_setopt($curlHandle, CURLOPT_VERBOSE, true);
+        $curlHeaders = [];
+        foreach ($context->headers as $key => $value) {
+            array_push($curlHeaders, $key . ": " . $value);
```
maxiloc: I rather use ```$curlHeaders[] = $key . ": " . $value```

We should probably make sure their is no duplicates. Otherwise good to go

Nagriar: For the duplicate it's ok since we convert a key => val to a X => key: val.

PR #77, opened 2016-05-06
Randomize the fallback hosts on a per client instance basis.

### LINKED ISSUE #75: Implement the new API client spec
APPID-1.algolianet.com, APPID-2.algolianet.com and APPID-3.algolianet.com need to be in random order

rayrutjes:
Fixes #75

### DIFF
--- a/tests/AlgoliaSearch/Tests/ClientContextTest.php
+++ b/tests/AlgoliaSearch/Tests/ClientContextTest.php
@@ -1,0 +1,101 @@
+<?php
+
+namespace AlgoliaSearch\Tests;
+
+use AlgoliaSearch\ClientContext;
+
+class ClientContextTest extends \PHPUnit_Framework_TestCase
+{
+    public function testRandomReadFallbackHosts()
+    {
+        $context = new ClientContext('whatever', 'whatever', null);
+        $hosts = $context->readHostsArray;
+
+        // Here we check that different calls results in the hosts being in and different order.
+        $isRandom = false;
+        for ($i = 0; $i < 100; $i++) {
+            $otherContext = new ClientContext('whatever', 'whatever', null);
+            if ($hosts !== $otherContext->readHostsArray) {
+                $isRandom = true;
+                break;
+            }
+        }
+
+        $this->assertTrue($isRandom);
+
+        // Check that the first entry is the correct API endpoint.
+        $this->assertEquals('whatever-dsn.algolia.net', $hosts[0]);
+
+        // As hosts are in a random order, we sort everything to be sure the correct hosts are present.
+        sort($hosts, SORT_STRING);
+        $expectedHosts = [
+            'whatever-1.algolianet.com',
+            'whatever-2.algolianet.com',
+            'whatever-3.algolianet.com',
+            'whatever-dsn.algolia.net',
+        ];
+        $this->assertEquals($expectedHosts, $hosts);
+    }
+
+    public function testRandomReadPlacesFallbackHosts()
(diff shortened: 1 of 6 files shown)


rayrutjes:
I guess we are gonna need to either remove the security tests or refactor everything to make sure tests passes when they really should.

tests/AlgoliaSearch/Tests/InitPlacesTest.php
```diff
@@ -9,13 +9,7 @@ class InitPlacesTest extends AlgoliaSearchTestCase
...
     {
         $placesIndex = Client::initPlaces(getenv('ALGOLIA_APPLICATION_ID'), getenv('ALGOLIA_API_KEY'));
-
-        $this->assertEquals([
-            'places-dsn.algolia.net',
-            'places-1.algolianet.com',
-            'places-2.algolianet.com',
-            'places-3.algolianet.com'
-        ], $placesIndex->getContext()->readHostsArray);
+        $this->assertInstanceOf('\AlgoliaSearch\PlacesIndex', $placesIndex);
```
maxiloc: We can not do equals anymore but maybe contains ?

rayrutjes: The ClientContext is fully tested in this very PR, including the places hostnames. Here we are testing initPlaces behaviour so we just have to ensure that the instance is the expected one. It would be redundant to re-test the hostnames here.

tests/AlgoliaSearch/Tests/ClientContextTest.php
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public function testRandomReadPlacesFallbackHosts()
+    {
+        $context = new ClientContext('whatever', 'whatever', null, true);
+        $hosts = $context->readHostsArray;
+
+        // Here we check that different calls results in the hosts being in and different order.
+        $isRandom = false;
+        for ($i = 0; $i < 100; $i++) {
```
maxiloc: This works but I don't like it 😄. Mocking the shuffle method would be better, but it adds complexity.
Not sure what's the best way to tackle this one

tests/AlgoliaSearch/Tests/InitPlacesTest.php
maxiloc: indeed 👍

tests/AlgoliaSearch/Tests/ClientContextTest.php
rayrutjes: Well if we mock the shuffle method we remove the behaviour we want to actually test. Maybe I can improve the readability of the test. Digging into it right now.

rayrutjes: After some readings, I realise testing for randomness is tricky. I think the actual tests provide us with a good first level. The code not being that complicated either, I think we can leave it as is.

maxiloc: Yes good for me. It works :). There is just a really really really low probability that the test fails.

rayrutjes: You are definitely right about that ;)

rayrutjes: Actually if the tests fails it means that the distribution is not very random so we are actually fine. Good to merge?

maxiloc: LGTM

PR #79, opened 2016-05-09
Fix security tests

JanPetr:


tests/AlgoliaSearch/Tests/SecurityTest.php
```diff
@@ -32,47 +36,102 @@ protected function tearDown()
...
         $this->assertTrue($newKey['key'] != '');
-        $resAfter = $this->index->listUserKeys();
-        $this->assertTrue($this->containsValue($resAfter['keys'], 'value', $newKey['key']));
         $this->assertFalse($this->containsValue($res['keys'], 'value', $newKey['key']));
-        $key = $this->index->getUserKeyACL($newKey['key']);
+
+        $asserted = false;
+        $timeouted = time() + $timeout;
+        while (time() < $timeouted) {
+            usleep(1000);
```
rayrutjes: Actually each loop is triggering about 300 calls to the api, we could increase the wait time by *1000.

tests/AlgoliaSearch/Tests/SecurityTest.php
```diff
@@ -32,47 +36,102 @@ protected function tearDown()
...
-        sleep(5);
+
         $this->assertTrue($newKey['key'] != '');
-        $resAfter = $this->index->listUserKeys();
-        $this->assertTrue($this->containsValue($resAfter['keys'], 'value', $newKey['key']));
         $this->assertFalse($this->containsValue($res['keys'], 'value', $newKey['key']));
-        $key = $this->index->getUserKeyACL($newKey['key']);
+
+        $asserted = false;
+        $timeouted = time() + $timeout;
```
maxiloc: Looks good. We can probably abstract this timeout logic. What do you think ?

JanPetr: Yes, I think it would be best but I don't know exactly how to do it as the content of while cycles is always different :-/

maxiloc: No I think the loop can go in a function that will take a callback as parameter

JanPetr: Yes, I realized that a minute ago. I just updated my comment. Will do ;)

maxiloc:
@JanPetr Can you rebase your commits ?

PR #91, opened 2016-05-15
Added missing class loadings.

rayrutjes:


rayrutjes:
@maxiloc please confirm that there is no reasons to not add the new classes.

maxiloc:
No problem. But not sure this file as a use

PR #94, opened 2016-05-16
Fix AccessTest to be resistant against DSN attack

JanPetr:


tests/AlgoliaSearch/Tests/AccessTest.php
```diff
@@ -8,20 +8,31 @@ class AccessTest extends AlgoliaSearchTestCase
 {
     public function testHTTPAccess()
     {
-        $client = new Client(getenv('ALGOLIA_APPLICATION_ID'), getenv('ALGOLIA_API_KEY'), ['http://'.getenv('ALGOLIA_APPLICATION_ID').'-1.algolia.io']);
+        $client = new Client(getenv('ALGOLIA_APPLICATION_ID'), getenv('ALGOLIA_API_KEY'), [
```
rayrutjes: The Client should be resilient without any hosts as argument.

JanPetr: Client itself is. But AccessTests are not. If you want to test this simple stuff (HTTP, HTTPS, Options, ...) you have to pass hosts manually. And as in the tests was only one domain (pretty old actually), the tests failed on DDos for example.
That's why I updated those domains and put there fallback domains.

maxiloc:
👍

rayrutjes:
I still do not get why we pass the hosts as argument, this is handled by ClientContext.
What is wrong with:
```php
public function testAccessWithOptions()
    {
        $client = new Client(
            getenv('ALGOLIA_APPLICATION_ID'), 
            getenv('ALGOLIA_API_KEY'), 
            null, 
            ['curloptions' => ['CURLOPT_FAILONERROR' => 0]]
        );

        $client->isAlive();
    }
``` 
If we want to test access, we should rather do it with built in behaviour instead of test specific hosts.
What do you think?

JanPetr:
Yes, that's true that the last test can be run without specified hosts. Other tests still test HTTP/HTTPS and that cannot be done otherwise. The question is if we need such tests. What do you think @maxiloc?
I would just update `testAccessWithOptions` to use default hosts and merge, hm?

maxiloc:
Yes we can go with @rayrutjes suggestion, I think it makes more sense.

JanPetr:
OK, updated, tests pass ... OK to merge @rayrutjes?

rayrutjes:
From my POV we don't need read hosts at all. We are not testing here that the read hosts are well passed to the underlying transport, we are just making a basic http request. What value does it add to override the default read hosts, do I miss something here?

JanPetr:
Here you are passing the protocol, so you can check if it works either on HTTP or HTTPS. This is what you cannot do when you do not pass hosts explicitly. That's why we have this test.

rayrutjes:
Ok then, I did miss the httpS diff. By the way, is there a valid reason to not use https?
LGTM

PR #103, opened 2016-05-23
Check AsynchDNS

JanPetr:


src/AlgoliaSearch/Client.php
```diff
@@ -1042,4 +1047,20 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
 
         return $client->getPlacesIndex();
     }
+
+    private function isAsynchDNSAvailable()
+    {
+        if (getenv('TRAVIS') == true) {
+            return true;
```
maxiloc: Should be false non ?

src/AlgoliaSearch/Client.php
```diff
@@ -1042,4 +1047,20 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
 
         return $client->getPlacesIndex();
     }
+
+    private function isAsynchDNSAvailable()
+    {
+        if (getenv('TRAVIS') == true) {
```
rayrutjes: Not comfortable with calling phpinfo like this. Also with should take care to not break existing implementations.

src/AlgoliaSearch/Client.php
```diff
@@ -1042,4 +1047,20 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
...
+            return true;
+        }
+
+        ob_start();
+        phpinfo();
```
rayrutjes: Not comfortable with calling phpinfo like this. Also we should take care to not break existing implementations.

src/AlgoliaSearch/Client.php
```diff
@@ -1042,4 +1047,20 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
...
+        $phpInfoString = ob_get_clean();
+
+        $pattern = '@.*<td class="e">AsynchDNS </td><td class="v">(.*) </td>.*@';
```
rayrutjes: What if for some reason the output differes even a bit from one environment to another?

rayrutjes:
Even if we have narrowed down the issue, we should take care not impacting existing users. 
We should probably raise a notices at most.
We should try to find a way to avoid the output capturing/phpinfo/regex/travis detection.

That said, well done for tracking down the source of the timeout problems on travis!

src/AlgoliaSearch/Client.php
maxiloc: Me neither but this is the only way, the info is not at another place

src/AlgoliaSearch/Client.php
maxiloc: Same here, we don't have any other way

src/AlgoliaSearch/Client.php
JanPetr: If it would be false, the warning will be triggered on Travis.

src/AlgoliaSearch/Client.php
JanPetr: There is no other way how to detect if AsynchDNS is available besides capturing phpinfo()

src/AlgoliaSearch/Client.php
JanPetr: We can try all the PHP version, compare their output and modify regex to handle all of them. Cannot see any other option :(

JanPetr:
> We should probably raise a notices at most.

Agree, it will be less ofensive than warning and it may cause less issues.

> to avoid ~~the output capturing/phpinfo/regex/~~travis detection.

Agree on Travis detection, it should be only in tests. I will refactor it tomorrow.

src/AlgoliaSearch/Client.php
JanPetr: Anyway, the condition should not be here at all. It should be only in tests. I will refactor it.

rayrutjes:
Maybe we should only make the tests pass if curl is not compiled with asynchdns, and forget about the rest. This way we take no risk.
We have to document this though.
The day we bump to 2.0.0 we can introduce this requirement check. 

Still a bit scared of what impact it would have to put such a check on every client instantiation

What do you think?

JanPetr:
> Maybe we should only make the tests pass if curl is not compiled with asynchdns, and forget about the rest.

That I do not agree. We have to somehow tell users that it might cause issues with DNS failover as it might effect our SLAs and support. But we should give them easy option how to supress the warning/notice. For example fifth parameter of constructor which will say "Do not check AsynchDNS", but it's default value will be to check it.

JanPetr:
And of course I would release it as 1.X version and not 1.9.X ...

rayrutjes:
Breaking their existing implementation is not an option either, that is why I suggest taking our precautions.

If we go down the road of integrating this into a non major release we should:
- ensure that phpinfo is always outputed with the same formatting regardless of the environment
- ensure that calling phpinfo does not impact performances and is always available
- ensure that curl info about AsynchDNS is trustworthy in phpinfo and always available
- ensure that there is no other legit way to handle timeout properly with curl
- not rely on any TRAVIS related environment variable if we don't want to end up with brittle tests. Actually, this PR does not pass a single test on my side even if it passes all the tests on travis.

Also we should test this implementation properly because actually it breaks clients which have not the guzzle AsynchDNS extension with:
`Undefined offset: 1` because of `$regs[1]` not being filled if `preg_match` does not match anything.

In either case we need to properly document this, as I agree that we should guide our users towards an environment that supports timeouts.

JanPetr:
Hey guys,
so I made couple of changes.

- Fixed failing tests on local because of different output of `phpinfo()` in CLI
- Added test to check DNS fallback
- Mesuared the time cost of check - the avarage time cost is `0.000916` seconds, capturing `phpinfo()` and `regex` included. Average is after 5000 loops.
- Changed `warning` into less aggressive `notice`
- Added mechanism to simply suppress `notice` as you can pass 5th attribute to `Client` class constructor

I still need to check all the PHP versions and it's `phpinfo()` output to make sure that `preg_match` is working correctly on all versions.

What I want to discuss with you is `TRAVIS` condition. It is little bit tricky as `Client` class is created withing the tests couple of times. If we do not want the condition to be there https://github.com/algolia/algoliasearch-client-php/pull/103/commits/fc599ef949ef22561f78a7baaa353e092ee97bcd#diff-795a3b6ca232a62a0ff6400631863703R1055 we will have to update all `Client` class creation in tests to pass 5th parameter to it's constructor as `!getenv('TRAVIS')`.

What do you say? Is it OK from your POV? Or do you have any other ideas?

src/AlgoliaSearch/Client.php
```diff
@@ -1042,4 +1048,29 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
 
         return $client->getPlacesIndex();
     }
+
+    private function isAsynchDNSAvailable()
+    {
+        // Do not run this test on Travis-CI as there is AsynchDNS disabled
+        if (getenv('TRAVIS') == true) {
```
rayrutjes: This is not useful here, given that we have the check in the test already.

src/AlgoliaSearch/Client.php
```diff
@@ -1042,4 +1048,29 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
...
+            return true;
+        }
+
+        for ($i = 0; $i < 5000; $i++) {
```
rayrutjes: Think you forgot to remove the benchmark loop ;)

src/AlgoliaSearch/Client.php
JanPetr: Did you read my notice above?

> What I want to discuss with you is TRAVIS condition. It is little bit tricky as Client class is created withing the tests couple of times. If we do not want the condition to be there fc599ef#diff-795a3b6ca232a62a0ff6400631863703R1055 we will have to update all Client class creation in tests to pass 5th parameter to it's constructor as !getenv('TRAVIS').

src/AlgoliaSearch/Client.php
```diff
@@ -1042,4 +1048,29 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
...
+            ob_start();
+            phpinfo();
+            $phpInfoString = ob_get_clean();
+
+            $pattern = '@.*<td class="e">AsynchDNS </td><td class="v">(.*) </td>.*@';
+            if (php_sapi_name() === 'cli') {
+                $pattern = '@.*AsynchDNS => (.*).*@';
+            }
+
+            if (preg_match($pattern, $phpInfoString, $regs) && false) {
```
rayrutjes: `* && false` always equals false right?

JanPetr: Oh no, my bad. It was there on testing puposes.

src/AlgoliaSearch/Client.php
rayrutjes: You mean notices make travis fail?

JanPetr: That's good question, I automaticaly assumed it would make Travis fail. I'll check it.

PR #109, opened 2016-05-24
Check asynchDNS

### LINKED ISSUE #105: Trigger warning when cURL AsynchDNS is not enabled

JanPetr:
New PR from #103, because of new target branch.

Fixes #105

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -70,16 +70,22 @@ class Client
      * @param string     $apiKey        a valid API key for the service
      * @param array|null $hostsArray    the list of hosts that you have received for the service
      * @param array      $options
+     * @param bool       $checkAsynchDNS
      *
      * @throws \Exception
      */
-    public function __construct($applicationID, $apiKey, $hostsArray = null, $options = array())
+    public function __construct($applicationID, $apiKey, $hostsArray = null, $options = array(), $checkAsynchDNS = true)
     {
         if (!function_exists('curl_init')) {
             throw new \Exception('AlgoliaSearch requires the CURL PHP extension.');
         }
+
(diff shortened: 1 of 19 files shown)


JanPetr:
Method `isAsynchDNSAvailable` tested on all PHP >= 5.4. Both CLI and HTML output. Works on all the versions 💯

rayrutjes:
The work done here should be bundled in a small script that allows to check the requirements.
Some kind of "benchmark" script would be great.

I'm still not convinced of the reliability of this, and we can not rely on a function that is blocked for security reasons in some environments.

src/AlgoliaSearch/Client.php
```diff
@@ -1032,14 +1038,39 @@ private function getPlacesIndex()
...
+    /**
+     * Checks if the server has enabled AsynchDNS feature
+     * The feature is required for correct DNS fallbacks
+     *
+     * @return bool
+     */
+    private function isAsynchDNSAvailable()
+    {
+        ob_start();
+        phpinfo();
```
lt: `curl.h` contains
```c
#define CURL_VERSION_ASYNCHDNS    (1<<7)  /* Asynchronous DNS resolves */
```

In theory something like this should work, but I haven't tested.

```php
$curlVersion = curl_version();
$hasAsynchDNS = $curlVersion['features'] & (1 << 7);
```

rayrutjes: Thanks for that @lt. 
We will test this the moment we decide to move on with this detection.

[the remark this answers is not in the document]
julienbourdeau: Thanks @lt for your inputs 🙏 very clever! 
I opened https://github.com/algolia/algoliasearch-client-php/pull/464 to port this to the client v2.

PR #112, opened 2016-06-06
Make deleteByQuery based on browse feature instead of search. Fixes #107

### LINKED ISSUE #107: Make deleteByQuery based on browse feature instead of search

maxiloc:
Fixes #107

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -390,19 +390,39 @@ class Index
         $args['distinct'] = false;
 
         $deletedCount = 0;
-        $results = $this->search($query, $args);
-        while ($results['nbHits'] != 0) {
-            $objectIDs = array();
-            foreach ($results['hits'] as $elt) {
-                array_push($objectIDs, $elt['objectID']);
+        $objectIDs = array();
+
+        $res = null;
+        while (true) {
+            $cursor = null;
+            $break = false;
+            do {
+                $iterator = $this->browseFrom($query, $args, $cursor);
+
+                if (count($iterator['hits']) == 0) {
+                    $break = true;
+                    break;
+                }
+
+                foreach ($iterator['hits'] as $hit) {
+                    $objectIDs[] = $hit['objectID'];
+                    $deletedCount++;
+                }
+
+                $cursor = isset($iterator['cursor']) && $iterator['cursor'] ? $iterator['cursor'] : null;
+            } while ($cursor !== null);
+
+            foreach (array_chunk($objectIDs, 1000) as $chunk) {
+                $res = $this->deleteObjects($chunk);
             }
-            $res = $this->deleteObjects($objectIDs);
-            $deletedCount += count($objectIDs);
-            if ($results['nbHits'] < $args['hitsPerPage'] && false === $waitLastCall) {
+
+            if ($waitLastCall && $res !== null) {
+                $this->waitTask($res['taskID']);
+            }
+
+            if ($break) {
                 break;
             }
-            $this->waitTask($res['taskID']);
-            $results = $this->search($query, $args);
         }
 
         return $deletedCount;


src/AlgoliaSearch/Index.php
```diff
@@ -390,19 +390,19 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
...
+        foreach ($this->browse($query, $args) as $hit) {
+            $objectIDs[] = $hit['objectID'];
+            $deletedCount++;
+        }
+
+        foreach (array_chunk($objectIDs, 1000) as $chunk) {
+            $res = $this->deleteObjects($chunk);
+
+            if ($waitLastCall) {
+                $this->waitTask($res['taskID']);
```
JanPetr: Now you are waiting for every call, not only last one, no? Is it intention?

maxiloc: no 😄

Nagriar:
seems good for me 👍

maxiloc:
@JanPetr Good ?

src/AlgoliaSearch/Index.php
```diff
@@ -390,19 +390,33 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
...
+        $objectIDs = [];
+
+        $res = null;
+        while (true) {
+            $cursor = null;
+            do {
+                $iterator = $this->browseFrom($query, $args, $cursor);
+
+                if (count($iterator['hits']) == 0) {
+                    break 2;
```
JanPetr: Once you get 0 records, you will break `while (true)` loop and you will never get to the `foreach` where you call `deleteObjects`, no? Am I missing something?

JanPetr: Or when I am on the last page, the `cursor` is not set? :)

maxiloc: no it's not set, yes there is a issue with the break

maxiloc:
@JanPetr I fixed the breaking issue

src/AlgoliaSearch/Index.php
JanPetr: And do we really need 'while(true)'?

On Monday, 6 June 2016, maxiloc <notifications@github.com> wrote:

> In src/AlgoliaSearch/Index.php
> <https://github.com/algolia/algoliasearch-client-php/pull/112#discussion_r65916163>
> :
>
> > @@ -390,19 +390,33 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
> >          $args['distinct'] = false;
> >
> >          $deletedCount = 0;
> > -        $results = $this->search($query, $args);
> > -        while ($results['nbHits'] != 0) {
> > -            $objectIDs = array();
> > -            foreach ($results['hits'] as $elt) {
> > -                array_push($objectIDs, $elt['objectID']);
> > +        $objectIDs = [];
> > +
> > +        $res = null;
> > +        while (true) {
> > +            $cursor = null;
> > +            do {
> > +                $iterator = $this->browseFrom($query, $args, $cursor);
> > +
> > +                if (count($iterator['hits']) == 0) {
> > +                    break 2;
>
> no it's not set, yes there is a issue with the break
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/algolia/algoliasearch-client-php/pull/112/files/0069823dcc9f6effb92576613a19ebee07a73494#r65916163>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABAlYA3WxCh0cl0KxYesjSN1E5jYSlC7ks5qJEI_gaJpZM4Iu6-l>
> .
>


-- 
--

<https://www.algolia.com/>

  <https://twitter.com/JanPetr>  <https://www.linkedin.com/in/janpetr1>
<https://github.com/JanPetr>

Jan Petr
Software engineer

Phone: +33 6 47 18 77 61 <+33+6+47+18+77+61>
Website: algolia.com <https://www.algolia.com/#millisecond-matter>
<https://www.algolia.com/#milliseconds-matter>

We are hiring! <https://www.algolia.com/jobs#milliseconds-matter>

maxiloc: Yes just in case you are pushing data during the browse. @Nagriar In the end do we want to remove records that were not here when deleteByQuery was called ?

src/AlgoliaSearch/Index.php
```diff
@@ -390,19 +390,39 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
         $args['distinct'] = false;
 
         $deletedCount = 0;
-        $results = $this->search($query, $args);
-        while ($results['nbHits'] != 0) {
-            $objectIDs = array();
-            foreach ($results['hits'] as $elt) {
-                array_push($objectIDs, $elt['objectID']);
+        $objectIDs = [];
```
JanPetr: Tests are failing on 5.3. We support it again 😄

src/AlgoliaSearch/Index.php
```diff
@@ -390,19 +390,39 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
...
             }
-            $res = $this->deleteObjects($objectIDs);
-            $deletedCount += count($objectIDs);
-            if ($results['nbHits'] < $args['hitsPerPage'] && false === $waitLastCall) {
+
+            if ($waitLastCall && $res !== null) {
+                $this->waitTask($res['taskID']);
+            }
+
+            if ($break) {
```
JanPetr: The `while(true)` still does not make sense to me. You will always break it in the first iteration, no?

In `do while` you iterate all browse results. Once you got to the very last page where are no results, you break `do while`, run deleting `foreach` and break anyway. No matter if there were another records pushed in meanwhile. Or do I get it wrong? 😕

Nagriar:
@JanPetr The issue is when you use the browse method in parallel of indexing operation, the browse isn't exact so you can miss record because the index get optimized or the position of the record moved between two calls. 
We need to keep this loop also the deleteByQuery doesn't clean the potential pending records that are not indexed so I think it's ok to delete record added during the run of the method.

JanPetr:
@Nagriar Yes, I understand why it's there. I just think that it will never go to another cycle than first.

rayrutjes:
Not sure I get the full picture here but I am just worried about one thing, is that we should ensure the results returned by the search are the same as the browse, and that they both deal with the ongoing tasks in the same fashion. I they do not we might break actual implementations beginning with mines.

It would be nice if we could write down the fundamental differences between the browse and the search API feature so that every one can measure whenever one or another is a good fit.

To me, the browse definitely made sense in the context of deleteByQuery but I am starting to doubt.

JanPetr:
I take back everything I said about `while(true)`. I tested it and on the last page there is no cursor returned from API. I didn't know that and I tought we will request API until we get 0 results (from non-existing page).

Now it's OK and works perfectly :)

Just fix the 5.3 compatibility pls @maxiloc :)

src/AlgoliaSearch/Index.php
```diff
@@ -390,19 +390,39 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
...
+        $objectIDs = [];
+
+        $res = null;
+        while (true) {
+            $cursor = null;
+            $break = false;
+            do {
+                $iterator = $this->browseFrom($query, $args, $cursor);
+
+                if (count($iterator['hits']) == 0) {
```
JanPetr: Wouldn't be better to use `$iterator['nbHits']` instead of counting?

maxiloc: I am not sure iterator['nbHits'] is the same than the hits count

maxiloc:
Should be good for 5.3

JanPetr:
@maxiloc Do you mind changing the target branch to `v2.0`? Or do you want to release it before the `2.0` major?

maxiloc:
2.0 is better

JanPetr:
ok, will change the target

PR #114, opened 2016-06-14
JSON validations

### LINKED ISSUE #113: Check for JSON errors after json_encode() as well as json_decode()
If you attempt to send malformed UTF-8 to Algolia you get 400 Bad Request as the HTTP response, this is because `json_encode()` fails earlier and sends an empty string instead of the correct JSON packet.

There is already a sanity check for the returned JSON, it would be good to add a similar sanity check before attempting to send any JSON to make this situation easier to debug.

JanPetr:
Fixes #113

### DIFF
--- a/algoliasearch.php
+++ b/algoliasearch.php
@@ -32,3 +32,4 @@ require_once 'src/AlgoliaSearch/IndexBrowser.php';
 require_once 'src/AlgoliaSearch/PlacesIndex.php';
 require_once 'src/AlgoliaSearch/SynonymType.php';
 require_once 'src/AlgoliaSearch/Version.php';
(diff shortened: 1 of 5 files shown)


rayrutjes:
Refactoring JSON related operations to a class is a good idea.

- Why all static? 
- We should add tests, but I suppose you are working on that ;)

JanPetr:
> Why all static?

Because I don't think it something we need to create a new class for that. Static makes more sense for me for this simple functionality. Making it non-static could bring little bit more complexity and I cannot see any profit from it.

> We should add tests, but I suppose you are working on that ;)

Yes, tests are comming.

rayrutjes:
LGTM
In V2 we should make this JSON class non static and inject it properly into the Client

Issue #104, opened 2016-05-24
Drop support of officialy unsupported PHP versions


JanPetr:
- [x] 5.3

redox:
I thought we wanted to keep that for Wordpress where 21% of the users are still running 5.3, did I miss something?

JanPetr:
We discussed with guys that in 2.0 major we will drop support of 5.3. I will check with @rayrutjes how he plans to handle WP.

JanPetr:
So we will not drop PHP 5.3 support until we release new WP plugin and have relevant data about PHP versions used on WP instances. 

I'm closing the issue now.

PR #116, opened 2016-06-14
Drop PHP 5.3 support

JanPetr:
Implements #104


JanPetr:
We will not drop PHP 5.3 support until we release new WP plugin and have relevant data about PHP versions used on WP instances.

PR #125, opened 2016-07-08
Send strategy as POST attribute and not as part of URL

### LINKED ISSUE #124: Put strategy in the POST body and not in the url
https://github.com/algolia/algoliasearch-client-php/blob/fc7502fc5cadddba100f40cd331206ee267f49bb/src/AlgoliaSearch/Client.php#L251

JanPetr:
Fixes #124


maxiloc:
👍, I would be more confident with a test

JanPetr:
> I would be more confident with a test

Added

PR #135, opened 2016-07-25
Add some new exceptions

modnarlluf:
Hi.

I added some new exceptions to make error handling easier. We have a project where we need some specific processing depending on the error returned by the API, and it would be nice if we can have it thrown directly by the client api. :) 

(I didn't move the AlgoliaException.php to keep compatibility.)

### DIFF
--- a/algoliasearch.php
+++ b/algoliasearch.php
@@ -25,8 +25,11 @@
  */
 
 require_once 'src/AlgoliaSearch/AlgoliaException.php';
+require_once 'src/AlgoliaSearch/Exception/AlgoliaBatchException.php';
+require_once 'src/AlgoliaSearch/Exception/AlgoliaDisjunctiveFacetsInvalidException.php';
 require_once 'src/AlgoliaSearch/Exception/AlgoliaIndexNotFoundException.php';
 require_once 'src/AlgoliaSearch/Exception/AlgoliaRecordTooBigException.php';
+require_once 'src/AlgoliaSearch/Exception/AlgoliaRefinementsInvalidException.php';
 require_once 'src/AlgoliaSearch/Client.php';
 require_once 'src/AlgoliaSearch/ClientContext.php';
 require_once 'src/AlgoliaSearch/Index.php';
--- a/algoliasearch.php
+++ b/algoliasearch.php
@@ -25,6 +25,8 @@
  */
 
 require_once 'src/AlgoliaSearch/AlgoliaException.php';
+require_once 'src/AlgoliaSearch/Exception/AlgoliaIndexNotFoundException.php';
+require_once 'src/AlgoliaSearch/Exception/AlgoliaRecordTooBigException.php';
 require_once 'src/AlgoliaSearch/Client.php';
 require_once 'src/AlgoliaSearch/ClientContext.php';
 require_once 'src/AlgoliaSearch/Index.php';
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -27,6 +27,9 @@
 
 namespace AlgoliaSearch;
 
+use AlgoliaSearch\Exception\AlgoliaRecordTooBigException;
+use AlgoliaSearch\Exception\AlgoliaIndexNotFoundException;
+
 /**
  * Entry point in the PHP API.
  * You should instantiate a Client object with your ApplicationID, ApiKey and Hosts
@@ -914,7 +917,7 @@ class Client
         curl_close($curlHandle);
 
         if (intval($http_status / 100) == 4) {
-            throw new AlgoliaException(isset($answer['message']) ? $answer['message'] : $http_status + ' error');
+            throw $this->throwCorrectException($answer, $http_status, $data);
         } elseif (intval($http_status / 100) != 2) {
             throw new \Exception($http_status.': '.$response);
         }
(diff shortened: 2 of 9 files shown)


algoliasearch.php
```diff
@@ -25,6 +25,8 @@
  */
 
 require_once 'src/AlgoliaSearch/AlgoliaException.php';
```
maxiloc: We can probably put this one in the Exception namespace also

maxiloc: But that would not be backward compatible, let's forget this

src/AlgoliaSearch/Client.php
```diff
@@ -1023,4 +1026,25 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
...
     }
+
+    /**
+     * @param $answer
+     * @param $http_status
+     * @param $data
+     *
+     * @return AlgoliaException|AlgoliaRecordTooBigException|AlgoliaIndexNotFoundException
+     */
+    protected function throwCorrectException($answer, $http_status, $data)
```
maxiloc: I think `throwException` or `throwAlgoliaException` makes more sense

algoliasearch.php
modnarlluf: Won't it create compatibility break for everyone catching AlgoliaException ? I know it's just to replace `use AlgoliaSearch\AlgoliaException;` to `use AlgoliaSearch\Exception\AlgoliaException;` but it can be annoying for some people.

maxiloc:
@modnarlluf Thanks for contributing.
Looks good for me, I added some comments.

@rayrutjes @JanPetr What's your opinion on this ?

algoliasearch.php
maxiloc: Yes, your totally right, just added a message after posting it to say exactly that

src/AlgoliaSearch/Client.php
modnarlluf: throwAlgoliaException sounds good to me

src/AlgoliaSearch/Client.php
```diff
@@ -914,7 +917,7 @@ public function doRequest(
         curl_close($curlHandle);
 
         if (intval($http_status / 100) == 4) {
-            throw new AlgoliaException(isset($answer['message']) ? $answer['message'] : $http_status + ' error');
+            throw $this->throwCorrectException($answer, $http_status, $data);
```
rayrutjes: Needs to be updated as well ;)

rayrutjes:
This makes sense and shouldn't break anything given the new exceptions extends the AlgoliaException.

@modnarlluf Would you mind adding some tests?

Maybe we could add more typed exceptions as well while we are at it.

src/AlgoliaSearch/Client.php
modnarlluf: My bad :)

src/AlgoliaSearch/Client.php
```diff
@@ -1023,4 +1026,25 @@ public static function initPlaces($appId, $apiKey, $hostsArray = null, $options
...
+    /**
+     * @param $answer
+     * @param $http_status
+     * @param $data
+     *
+     * @return AlgoliaException|AlgoliaRecordTooBigException|AlgoliaIndexNotFoundException
+     */
+    protected function throwAlgoliaException($answer, $http_status, $data)
+    {
+        switch (true) {
```
rayrutjes: Personally not a huge fan of the `switch(true)` trick. Makes it difficult to read. Maybe we could go with simple if statements?

modnarlluf: As you wish, I don't really have preference about that

rayrutjes:
After a second read I realize that there might be BC issues as we are changing the messages of the exceptions.

I'd go with only Type sugar keeping the exception classes but removing the bodies. That will allow to explicitly check for a specific exception whilst keeping BC.

WDYT?

modnarlluf:
You are right. In our case it, it would actually create BC.

Would it be good if we don't change the message, but keep the $record property of AlgoliaRecordTooBigException and add a new $indexName property to AlgoliaIndexNotFoundException ?

rayrutjes:
@modnarlluf, creating some nice getters for these new exceptions totally makes sense from a design point of view, wondering though if this will ever be used.

modnarlluf:
We actually need this on our project... or maybe we are not doing something right. I will try to explain:

We use Algolia to save products from several merchants. Some of them have like 10k products, so we divide those products into chuncks (with a size of 1024) and use `$index->saveObjects($chunk);`.

The issue is when we get a "Record is too big" error (ex: when a product got a really long description), we don't know which product is wrong. For now, we divide again the chunk into two smaller chunks (with a size of 512), and try again.  We repeat it until size == 1. And then we know which product is wrong, and we can trigger specific processing.

We thought about checking the `Json::encode($object)` size, but the size we get and the size provided by the Algolia's error message don't match:

```php
<?php

require_once '../algoliasearch.php';

$client = new \AlgoliaSearch\Client('foo', 'bar');
$index = $client->initIndex('foo');
try {
    $index->clearIndex();
} catch(Exception $e) {}

$contacts = json_decode(file_get_contents('../contacts.json'));
$contacts = array_slice($contacts, 0, count($contacts)/1.3);
$object = ['contacts' => $contacts];
$objects = [$object];

var_dump(mb_strlen(\AlgoliaSearch\Json::encode($object), '8bit'));

$index->addObjects($objects);

// Output:
// int(115240)
// PHP Fatal error:  [...] Record is too big size=114462 bytes. [...]
```

I know it's a tiny difference, but we don't find this *robust* (but maybe it's better than divide chunks... IDK). For now, we never saw a greater difference than 3kb.

Our real issue is to actually get the wrong(s) record(s) when we are using `Index::saveObjects($objects)`. It would be nice if Algolia could send us the record which produces the issue. Or if `Index::saveObjects`/`Index::addObjects` could actually throw an `AlgoliaRecordsTooBigException` with the list of wrong records, reachable by `AlgoliaRecordsTooBigException::getRecords()`. And why not both... 😶 

The dream usage would be something like:
```php
try {
    $index->saveObjects($products);
} catch(AlgoliaRecordsTooBigException $e) {
    $this->invalidate($e->getRecords);
}
```

I'm open to suggestions !💃

maxiloc:
@modnarlluf We are going to try make the engine return the objectID that failed, it would be better and less probable to break. I'll keep updated

maxiloc:
@modnarlluf We just changed the error message to include the failing objectID
You won't have the list because we stop the request after seeing the first bad object.

The record limit is at 10kb but we have some margin to allow some record to be a bit bigger.
I would recommend to apply a processing that brings your records around 10kb because having records > 10kb is usually bad for relevance. If every record you send is around 10kb you should not have any exception.

modnarlluf:
@maxiloc Thx for your update ! It will be easier to index larges chunks of records this way, and quickly extract the wrongs ones. Sounds good :) 

We already got an upgrade to 100kb for the record limit. We are aware that is not the best concerning performances, maybe we are gonna change it soon...

Thx again :)

rayrutjes:
Hi @modnarlluf ,

We plan on integrating more fine tuned exceptions in an upcoming major release.
We will definitely inspire our work on what was discussed here.

Thanks a lot for contributing ❤️

PR #140, opened 2016-08-02
Follow new UA convention

JanPetr:


src/AlgoliaSearch/Version.php
```diff
@@ -32,9 +32,10 @@ class Version
     const VALUE = '1.10.1';
 
     public static $custom_value = '';
+    public static $custom_value_prefix = '';
 
     public static function get()
     {
-        return self::VALUE.static::$custom_value;
+        return self::$custom_value_prefix . 'Algolia for PHP ('.self::VALUE.')' . static::$custom_value;
```
rayrutjes: This breaks BC as we didn't have Algolia for PHP part of this before.

src/AlgoliaSearch/Version.php
```diff
@@ -33,8 +33,32 @@ class Version
...
     public static $custom_value = '';
 
+    private static $prefixUserAgentSegments = '';
+    private static $suffixUserAgentSegments = '';
+
+    // Method untouched to keep backward compatibility
     public static function get()
     {
-        return self::VALUE.static::$custom_value;
+        return self::VALUE . static::$custom_value;
```
maxiloc: Why can't we just do a concatenation ?

JanPetr: ? We still do, there are just spaces for better readability.

maxiloc: The thing I was saying is I think we can do everything from the get method and don't need the others

JanPetr: - there is new UA convention, you can find it in API specifications. There should be parantheses around version numbers. It's impossible to do that in `get()` method with keeping with BC
- By the convention, the integration version should be before API client
- By convention `;` should be used as separator and version numbers should be in parantheses. It's always easier to keep this rules on one place (`Version` class) and use it's methods than keeping those rules in mind when dealing with all the integrations

maxiloc: I see no issue in changing the get method. It don't see how it could impact existing users.

JanPetr: Previously the method returned just version number. The probability is very low that somebody is using it but it's still a BC break. I wouldn't touch as there is no need to touch it.

maxiloc: ok makes sense

PR #150, opened 2016-09-04
Fixed docblock

yosmanyga:


src/AlgoliaSearch/Index.php
```diff
@@ -327,7 +327,7 @@ public function saveObjects($objects, $objectIDKey = 'objectID')
     /**
      * Delete an object from the index.
      *
-     * @param $objectID the unique identifier of object to delete
+     * @param string $objectID the unique identifier of object to delete
```
rayrutjes: Actually 'int|string' might be more correct because the engine accepts both.
Thanks for the contribution!

yosmanyga: Done.

maxiloc:
@yosmanyga Thanks for contributing

PR #159, opened 2016-09-28
Rename slaves to replicas and use new setting names in docs

JanPetr:


src/AlgoliaSearch/Index.php
```diff
@@ -741,7 +741,7 @@ public function clearIndex()
      *                        - attributesToSnippet**: (array of strings) default list of attributes to snippet alongside the
      *                        number of words to return (syntax is attributeName:nbWords). By default no snippet is computed.
      *                        If set to null, no snippet is computed.
-     *                        - attributesToIndex: (array of strings) the list of fields you want to index.
+     *                        - searchableAttributes (attributesToIndex): (array of strings) the list of fields you want to index.
```
rayrutjes: Maybe also replace 'attributesToIndex'

rayrutjes:
Just 1 comment, looks good to me otherwise 👍

src/AlgoliaSearch/Index.php
JanPetr: It was `attributesToIndex` before, I replaced it by `searchableAttributes` and put `attributesToIndex` into parentheses because of "backward compatibility" (to be clear to everybody who might be used to old naming)

rayrutjes: The maybe write `(formerly named 'attributesToIndex')`?

JanPetr: Makes sense, it's more descriptive. Pushed

rayrutjes:
LGTM!

PR #185, opened 2016-12-02
Implement new retry strategy

maxiloc:


JanPetr:
LGTM

src/AlgoliaSearch/Client.php
```diff
@@ -704,31 +704,32 @@ public static function buildQuery($args)
...
+     * @internal param int $connectTimeout
      */
     public function request(
         $context,
         $method,
         $path,
         $params,
         $data,
-        $hostsArray,
+        HostsHandler $hostsArray,
```
rayrutjes: This introduces a backward compatibility issue.

maxiloc: Yes we'll need to do a 2.0

maxiloc: Ok I'll find a way

maxiloc:
@redox @rayrutjes @JanPetr Can you have a look again, I made it BC

redox:
> @redox @rayrutjes @JanPetr Can you have a look again, I made it BC

The only issue I see is that using such `static` attribute is not compatible with having several API clients with different `application_id` and/or `hosts`; right?

maxiloc:
We need it static to be able to handle the strategy when you instanciate twice with the same app id and host. But we can make it in a way that handle different app id and host

rayrutjes:
I have a solution underway. Should be there in a few.

Issue #181, opened 2016-11-25
Optimize re-try strategy by updating last working DN in the context


rayrutjes:
Currently, we will always try to connect to the same, eventually non reachable, DN before moving to the next one.

When a domain is not reachable, we should reflect it in the ClientContext so that we don't introduce a delay of the amount of the timeout for every call to the API for the same process.

PR #186, opened 2016-12-06
Implement statefull connection retry strategy

rayrutjes:
Resolves: #181

This PR will rotate the hosts array at each failed connection attempt. 
ClientContext holds the state and the Client simply ask for a rotation by passing the "justUnresolvable" host as argument.

**Shared state between different instances of the Client:**
Given the current design of the Client, sharing state between different Client instances is not feasable. Also I don't think it is a big issue because most of the time, there will only be one instance of the client per PHP thread.

If for some reason a user wants to share the state between multiple instances of a the client, he could simply override the ClientContext on all the Client.
**Shared state between different PHP processes**
In PHP processes don't share state between them, every process spawning is an attempt to recover connection to the main host. In a future version of the API Client, we could create an abstraction around the state to share it via storage layer (ideally in memory k/v store), but for now this will be overkill and very messy to implement.

Let me know your thoughts @maxiloc @JanPetr @redox

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,23 @@
 language: php
 
-php:
-  - 5.3
-  - 5.4
-  - 5.5
-  - 5.6
-  - 7.0
-  - hhvm
-
-sudo: false
+matrix:
+  include:
+    - php: 5.3
+    - php: 5.4
+      dist: trusty
+      sudo: required
+    - php: 5.5
+      dist: trusty
+      sudo: required
+    - php: 5.6
+      dist: trusty
+      sudo: required
+    - php: 7.0
+      dist: trusty
+      sudo: required
+    - php: hhvm
+      dist: trusty
+      sudo: required
 
 cache:
   directories:
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -110,7 +110,7 @@ class Client
     }
 
     /**
-     * Change the default connect timeout of 2s to a custom value
+     * Change the default connect timeout of 1s to a custom value
      * (only useful if your server has a very slow connectivity to Algolia backend).
      *
      * @param int $connectTimeout the connection timeout
@@ -743,6 +743,9 @@ class Client
                 throw $e;
             } catch (\Exception $e) {
                 $exceptions[$host] = $e->getMessage();
+                if ($context instanceof ClientContext) {
+                    $context->rotateHosts($host);
+                }
             }
         }
         throw new AlgoliaException('Hosts unreachable: '.implode(',', $exceptions));
--- a/src/AlgoliaSearch/ClientContext.php
+++ b/src/AlgoliaSearch/ClientContext.php
@@ -88,8 +88,8 @@ class ClientContext
      */
     public function __construct($applicationID, $apiKey, $hostsArray, $placesEnabled = false)
     {
-        // connect timeout of 2s by default
-        $this->connectTimeout = 2;
+        // connect timeout of 1s by default
+        $this->connectTimeout = 1;
 
         // global timeout of 30s by default
         $this->readTimeout = 30;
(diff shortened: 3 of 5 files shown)


maxiloc:
Yes that works also,
two things:
- I think we need to handle "Shared state between different instances of the Client", it is feasible and I would not bet on people not doing it.

```php
foreach ($products as $product) {
  $client = new Client();
  $index = $client->initIndex();
  [...]
}
```

is probably something done more than we would like it to.

- when
  * -dns fails
  * -1 works
  * -1 fails
  the next request should be on -dns and not -1

rayrutjes:
> I think we need to handle "Shared state between different instances of the Client", it is feasible and I would not bet on people not doing it.

Unfortunately, given the current design, I don't think it is possible without introducing a BC.
We can't force a shared ClientContext between instances because currently they might have their own implems for that.


> the next request should be on -dns and not -1

That is the scenario you told me about and had a hard time remembering! Will implement it, thx.

rayrutjes:
Also regarding the tests I configured a Matrix so that everything above php 5.3 uses the trusty distrib.

For 5.3 I skip the re-try strategy test as we have no way to execute it.

maxiloc:
👍, last commit I did make it work without the BC break

maxiloc:
It does not handle the multi app id host, but should not cause any problem to implement it

rayrutjes:
IMHO your solution is a bit overkill. There is certainly no perfect solution though given the current design of the client.

I would militate for an update we are sure will negatively impact as few users as possible. The less changes we make here the better. 

Regarding the shared state between client instances, I'm not convinced by the 'static' approach. In general I'm against it because it often hides a design problem (which we clearly have). I wouldn't add more concepts on top of the current client. I'm sure your implementation will work fine in most cases, but I'm just afraid that if we have another design obstacle and if we need to adjust the implementation, it will be even harder than today.

Also, the 2 left points are minor in comparison of what we already fix with ensuring we keep the state inside of a client instance.

Regarding the fail -> work -> fail -> reset, this would be a big issue if we had a shared state between PHP processes, but as we haven't, users will only very rarely (never) benefit from this optimization. Not saying we shouldn't come up with a clean solution for a future release of the client though.

Regarding users that would Loop over `new Client()`, it sure must happen somewhere but I think it's more of a design issue on their end. In all integrations + frameworks packages we always have 1 client. Users tend to use configuration from a file, instantiate the client, and share it inside the current app. For me this is comparable to prepared statements with PDO ;) 
Again, this is definitely something we should address later on, but I see more cons than pros to do it now.

maxiloc:
> IMHO your solution is a bit overkill. There is certainly no perfect solution though given the current design of the client.

Your solution is definitely simpler

> I would militate for an update we are sure will negatively impact as few users as possible. The less changes we make here the better.

For me the main goal is to reduce dns impact for the most possible users. With that in mind I would prefer it to be more complex than less resilient.

> Regarding the shared state between client instances, I'm not convinced by the 'static' approach. In general I'm against it because it often hides a design problem (which we clearly have). I wouldn't add more concepts on top of the current client. I'm sure your implementation will work fine in most cases, but I'm just afraid that if we have another design obstacle and if we need to adjust the implementation, it will be even harder than today.

Agreed

> Also, the 2 left points are minor in comparison of what we already fix with ensuring we keep the state inside of a client instance.

> Regarding the fail -> work -> fail -> reset, this would be a big issue if we had a shared state between PHP processes, but as we haven't, users will only very rarely (never) benefit from this optimization. Not saying we shouldn't come up with a clean solution for a future release of the client though.

Yes not a big issue

> Regarding users that would Loop over new Client(), it sure must happen somewhere but I think it's more of a design issue on their end. In all integrations + frameworks packages we always have 1 client. Users tend to use configuration from a file, instantiate the client, and share it inside the current app. For me this is comparable to prepared statements with PDO ;)
Again, this is definitely something we should address later on, but I see more cons than pros to do it now.

I tend to disagree here, there is no reason why this should less resilient even if it's bad design.

rayrutjes:
> For me the main goal is to reduce dns impact for the most possible users. 

We have the same goal ;)
That being said if we end up introducing issues in existing applications because of this change, it is going to be worse. Truth is we can't say for sure the impact of your solution compared to mine. Sure thing is that mine is probably a bit less risky.

With that in mind I would prefer it to be more complex than less resilient.



> I tend to disagree here, there is no reason why this should less resilient even if it's bad design.

I would love to come up with a better solution to allow better resiliency even for those use cases, but for me it still is a design issue both on our side + the side of the user.

I will put more thoughts into it today :)

rayrutjes:
What is your take on this @JanPetr ?

rayrutjes:
See #187

PR #187, opened 2016-12-07
Statefull retry strategy with multi client support (2)

maxiloc:


maxiloc:
This is a proposal with the simplicity of #186 while having the resilience of #185.
We do need a static variable but it I think it's inevitable

src/AlgoliaSearch/ClientContext.php
```diff
@@ -107,6 +113,18 @@ public function __construct($applicationID, $apiKey, $hostsArray, $placesEnabled
             $this->writeHostsArray = $this->getDefaultWriteHosts();
         }
 
+        $i = 0;
+        while ($i <= count($this->readHostsArray) && in_array($this->readHostsArray[0], static::$failingHosts)) {
```
rayrutjes: This piece of logic should go in the rotateHosts method

src/AlgoliaSearch/ClientContext.php
```diff
@@ -248,4 +266,37 @@ public function setExtraHeader($key, $value)
     {
         $this->headers[$key] = $value;
     }
+
+    /**
+     * @param $host
+     */
+    public function addFailingHost($host)
+    {
+        if (! in_array($host, static::$failingHosts)) {
```
rayrutjes: As we want to enforce a unique variable that holds the failing hosts we should use `self` to disable late static binding.

src/AlgoliaSearch/ClientContext.php
```diff
@@ -248,4 +266,37 @@ public function setExtraHeader($key, $value)
...
+    /**
+     * The Client will call this when an attempt to reach a given host times out.
+     * If the host is first either in the readHosts or the the writeHosts, we
+     * rotate the array to ensure the next API call will be directly made with a working
+     * host. This mainly ensures we don't add the equivalent of the connection timeout value to each
+     * request to the API.
+     *
+     * @param string $host The host that was unavailable.
+     */
+    public function rotateHosts($host)
```
rayrutjes: Remove the `$host` parameter as we will rely on the static list of failed hosts.

src/AlgoliaSearch/ClientContext.php
```diff
@@ -248,4 +266,37 @@ public function setExtraHeader($key, $value)
...
+    {
+        if (array_search($host, $this->readHostsArray, true) === 0) {
```
rayrutjes: Use the strategy currently implemented in the constructor instead.

tests/AlgoliaSearch/Tests/AccessTest.php
```diff
@@ -32,4 +32,45 @@ public function testAccessWithOptions()
 
         $client->isAlive();
     }
+
+    private function getNewClient()
```
rayrutjes: The name is misleading. We should name it something like "newClientWithUnresolvableHost" or something like that. When I reviewed your code at first, it took me some time to realize that the "newClient" was actually configured to fail on first attempt.

tests/AlgoliaSearch/Tests/AccessTest.php
```diff
@@ -32,4 +32,45 @@ public function testAccessWithOptions()
...
+    {
+        return new Client(
+            getenv('ALGOLIA_APPLICATION_ID'),
+            getenv('ALGOLIA_API_KEY'),
+            array(
+                getenv('ALGOLIA_APPLICATION_ID') . '.algolia.biz', // Assuming .biz will always fail to resolve
```
rayrutjes: We don't assume here, it will always fail ;)

tests/AlgoliaSearch/Tests/AccessTest.php
```diff
@@ -32,4 +32,45 @@ public function testAccessWithOptions()
...
+                getenv('ALGOLIA_APPLICATION_ID') . '.algolia.net'
+            )
+        );
+    }
+
+    public function testStatefullRetryStrategy()
```
rayrutjes: Not sure this adds something to our tests, I think we can discard and keep the other one.

tests/AlgoliaSearch/Tests/ClientContextTest.php
```diff
@@ -98,4 +98,38 @@ public function testRandomWriteFallbackHosts()
...
+    public function testHostsCanBeRotated()
+    {
+        $context = new ClientContext('whatever', 'whatever', null);
+        $initialReadHosts = array('host1.com', 'shared-host.com', 'host3.com');
+        $initialWriteHosts = array('write-host1.com', 'shared-host.com', 'write-host3.com');
+        $context->readHostsArray = $initialReadHosts;
+        $context->writeHostsArray = $initialWriteHosts;
+
+        // Rotate read host.
+        $context->rotateHosts('host1.com');
```
rayrutjes: These tests needs to be re-written to take into account the new strategy.

maxiloc: After the changes I made the test still makes sense

src/AlgoliaSearch/Client.php
```diff
@@ -743,6 +743,9 @@ public function request(
                 throw $e;
             } catch (\Exception $e) {
                 $exceptions[$host] = $e->getMessage();
+                if ($context instanceof ClientContext) {
+                    $context->rotateHosts($host);
```
rayrutjes: Your previous implementation here was better. Explicit call to register failed host. BTW the method can be static.

src/AlgoliaSearch/ClientContext.php
```diff
@@ -248,4 +256,42 @@ public function setExtraHeader($key, $value)
     {
         $this->headers[$key] = $value;
     }
+
+    /**
+     * @param $host
+     */
+    public function addFailingHost($host)
```
rayrutjes: We can make this static.

src/AlgoliaSearch/ClientContext.php
```diff
@@ -248,4 +256,42 @@ public function setExtraHeader($key, $value)
...
+    /**
+     * This method is called to pass on failing hosts.
+     * If a failingHost is passed we add it to the failingHosts array
+     * If the host is first either in the failingHosts array, we
+     * rotate the array to ensure the next API call will be directly made with a working
+     * host. This mainly ensures we don't add the equivalent of the connection timeout value to each
+     * request to the API.
+     * @param null $failingHost
+     */
+    public function rotateHosts($failingHost = null)
```
rayrutjes: Remove parameter to force relying on the static state.

maxiloc: I need to keep it to add the failing host inside the rotate

rayrutjes: not if you explicitly call the addFailedHost like you did before which makes more sense from a design perspective. Keep the concerns separated so that we can actually register a host that failed without rotating. And then make the rotation based on the state of the static variable.

rayrutjes:
You should probably rebase on master, as it seems you integrated changes that are already commited.

tests/AlgoliaSearch/Tests/AccessTest.php
```diff
@@ -32,4 +32,72 @@ public function testAccessWithOptions()
...
+        $this->assertEquals(array(
+            getenv('ALGOLIA_APPLICATION_ID') . '.algolia.net',
+            'APP_ID_1' . '.algolia.biz', // .biz will always fail to resolve
+            getenv('ALGOLIA_APPLICATION_ID') . '.algolia.biz'
+        ), $client3->getContext()->readHostsArray);
+    }
+
+    public function testStatefullRetryStrategyForSeveralInstance()
+    {
+        $start = microtime(true);
```
rayrutjes: missing the php 5.3. skip here

src/AlgoliaSearch/ClientContext.php
```diff
@@ -248,4 +256,36 @@ public function setExtraHeader($key, $value)
...
         $this->headers[$key] = $value;
     }
+
+    /**
+     * @param $host
+     */
+    public static function addFailingHost($host)
+    {
+        if (! in_array($host, self::$failingHosts)) {
+            self::$failingHosts[] = $host;
```
JanPetr: Cannot we do it even for multiple appIds? Like

```php
self::$failingHosts[$appId][]  = $host;
```

and in `rotateHosts` check with corresponding appId?

JanPetr: Oh, yeah, right :) gg 👍

maxiloc:
@rayrutjes @JanPetr Are we good to merge ?

tests/AlgoliaSearch/Tests/ClientContextTest.php
```diff
@@ -98,4 +98,35 @@ public function testRandomWriteFallbackHosts()
...
+
+        $context->addFailingHost('shared-host.com');
+        $context->rotateHosts();
+        $this->assertEquals(array('host3.com', 'host1.com', 'shared-host.com'), $context->readHostsArray);
+        $this->assertEquals(array('write-host1.com', 'shared-host.com', 'write-host3.com'), $context->writeHostsArray);
+
+        $context->addFailingHost('write-host1.com');
+        $context->rotateHosts();
+        $this->assertEquals(array('host3.com', 'host1.com', 'shared-host.com'), $context->readHostsArray);
+        $this->assertEquals(array('write-host3.com', 'write-host1.com', 'shared-host.com'), $context->writeHostsArray);
```
rayrutjes: Could we add 1 last test here where we create a new instance of the ClientContext to ensure we unit test the "statefull" host ordering accross ClientContext instances?

maxiloc: Just added one, can you check it

rayrutjes:
Looks like we have a brittle test for `testGetSaveObjects`.

tests/AlgoliaSearch/Tests/ClientContextTest.php
rayrutjes: Perfect thanks.

rayrutjes:
Ok this is my bad, I re-enabled parallel tests the other day so now it conflicts.
Fixing it right now and re-launching the tests @maxiloc

JanPetr:
👍

PR #193, opened 2016-12-09
Add performance tests

rayrutjes:


rayrutjes:
Travis adding some overhead here which makes this PR not very useful.

PR #195, opened 2016-12-10
Keep state between processes

rayrutjes:
We abstracted an interface for the failed hosts retrieval so that users can easily choose their caching strategy based on what is best according to their environment.

By default we push the new File Based cache strategy, but if the tmp directory is not writable, we fallback to the in memory strategy.

User can provide an implementation of FailingHostsCache as a replacement of this strategy.

Because of the current design of the Client, the FailingHostsCache implementation needs to be provided at Client initialization because that is where the ClientContext is instantiated and during that instantiation we need to rotate the hosts.

Regarding the BC static to non static, given it has only been available in 1 version and was only used internally + wasn't even called statically, I think we are safe doing the change.

TODO:
- [x] Unit tests
- [x] Choose if we push the file based cache by default or if we keep the in memory strategy
- [x] Implement in memory cache invalidation for long running PHP processes  > 5min
- [x] See what we do about the static to public method signature of addFailingHost (BC) that was only introduced in last version

### DIFF
--- a/src/AlgoliaSearch/ClientContext.php
+++ b/src/AlgoliaSearch/ClientContext.php
@@ -77,21 +77,20 @@ class ClientContext
     public $connectTimeout;
 
     /**
-     * @var array
+     * @var FailingHostsCache
      */
-    private static $failingHosts = array();
+    private $failingHostsCache;
 
     /**
-     * ClientContext constructor.
-     *
-     * @param string $applicationID
-     * @param string $apiKey
-     * @param array  $hostsArray
-     * @param bool   $placesEnabled
+     * @param string            $applicationID
+     * @param string            $apiKey
+     * @param array             $hostsArray
+     * @param bool              $placesEnabled
+     * @param FailingHostsCache $failingHostsCache
      *
      * @throws Exception
      */
-    public function __construct($applicationID, $apiKey, $hostsArray, $placesEnabled = false)
+    public function __construct($applicationID, $apiKey, $hostsArray, $placesEnabled = false, FailingHostsCache $failingHostsCache = null)
     {
         // connect timeout of 1s by default
         $this->connectTimeout = 1;
@@ -113,8 +112,6 @@ class ClientContext
             $this->writeHostsArray = $this->getDefaultWriteHosts();
         }
 
-        $this->rotateHosts();
-
         if ($this->applicationID == null || mb_strlen($this->applicationID) == 0) {
             throw new Exception('AlgoliaSearch requires an applicationID.');
         }
@@ -129,6 +126,14 @@ class ClientContext
         $this->algoliaUserToken = null;
         $this->rateLimitAPIKey = null;
         $this->headers = array();
+
+        if ($failingHostsCache === null) {
+            $this->failingHostsCache = new InMemoryFailingHostsCache();
+        } else {
+            $this->failingHostsCache = $failingHostsCache;
+        }
+
+        $this->rotateHosts();
     }
 
     /**
@@ -258,15 +263,20 @@ class ClientContext
     }
 
     /**
-     * @param $host
+     * @param string $host
      */
-    public static function addFailingHost($host)
+    public function addFailingHost($host)
     {
-        if (!in_array($host, self::$failingHosts)) {
-            self::$failingHosts[] = $host;
-        }
+        $this->failingHostsCache->addFailingHost($host);
     }
 
+    /**
+     * @return FailingHostsCache
+     */
+    public function getFailingHostsCache()
+    {
+        return $this->failingHostsCache;
+    }
     /**
      * This method is called to pass on failing hosts.
      * If the host is first either in the failingHosts array, we
@@ -276,14 +286,15 @@ class ClientContext
      */
     public function rotateHosts()
     {
+        $failingHosts = $this->failingHostsCache->getFailingHosts();
         $i = 0;
-        while ($i <= count($this->readHostsArray) && in_array($this->readHostsArray[0], self::$failingHosts)) {
+        while ($i <= count($this->readHostsArray) && in_array($this->readHostsArray[0], $failingHosts)) {
             $i++;
             $this->readHostsArray[] = array_shift($this->readHostsArray);
         }
 
         $i = 0;
-        while ($i <= count($this->writeHostsArray) && in_array($this->writeHostsArray[0], self::$failingHosts)) {
+        while ($i <= count($this->writeHostsArray) && in_array($this->writeHostsArray[0], $failingHosts)) {
             $i++;
             $this->writeHostsArray[] = array_shift($this->writeHostsArray);
         }
--- a/src/AlgoliaSearch/FileFailingHostsCache.php
+++ b/src/AlgoliaSearch/FileFailingHostsCache.php
@@ -1,0 +1,192 @@
+<?php
+
+namespace AlgoliaSearch;
+
+class FileFailingHostsCache implements FailingHostsCache
+{
+    /**
+     * Timestamp key used in the JSON representation.
+     */
+    const TIMESTAMP = 'timestamp';
+
+    /**
+     * Failing hosts key used in the JSON representation.
+     */
+    const FAILING_HOSTS = 'failing_hosts';
+
+    /**
+     * @var string
+     */
+    private $failingHostsCacheFile;
+
+    /**
+     * @var int
+     */
+    private $ttl;
+
+    /**
+     * @param int|null    $ttl The time to live of the cache in seconds.
+     * @param string|null $file
+     *
+     */
+    public function __construct($ttl = null, $file = null)
+    {
+        if ($file === null) {
+            $this->failingHostsCacheFile = $this->getDefaultCacheFile();
+        } else {
+            $this->failingHostsCacheFile = (string) $file;
+        }
+
+        $this->assertCacheFileIsValid($this->failingHostsCacheFile);
+
+        if ($ttl === null) {
+            $ttl = 60 * 5; // 5 minutes
+        }
+
+        $this->ttl = (int) $ttl;
+    }
+
+    /**
+     * @return int
+     */
+    public function getTtl()
+    {
+        return $this->ttl;
+    }
+
+    /**
+     * @param $file
+     */
+    private function assertCacheFileIsValid($file)
+    {
+        $fileDirectory = dirname($file);
+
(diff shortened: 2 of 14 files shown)


maxiloc:
is it expected that the test are not passing with hhvm ?

maxiloc:
Good job on this. Looks good.
Still have some doubt on file first but except that we're good to go

rayrutjes:
I don't expect the tests to fail. I will make sure they pass on HHVM.
I see were it broke the other tests. Should be an easy fix.

rayrutjes:
I'm afraid it is an HHVM limitation: https://gist.github.com/tlevi/8503668
I will look for a workaround.

rayrutjes:
Here is a workaround: https://github.com/LExpress/symfony1/pull/127/files
Will see if there are other cleaner solutions.

rayrutjes:
What's left to be done:

- [ ] I will revert to in memory by default. We never know about the side-effects of the file based caching strategy and I don't want to take the risk to break existing implementations.
- [ ] I will more thoroughly test the file based strategy

maxiloc:
Good for me.
We'll probably need to mention this in the doc

JanPetr:
I'm really not sure about `eval()` in the file cache. Why did you decide to go this way?
`eval()` is not supported on vast majority of hostings and is considered to be security vulnerability. 

Why not just use text file with specific syntax which is easy to parse? The code will be cleaner, easier to read, we won't have to deal with OPCache and HHVM workaround.

rayrutjes:
`eval` is only used in HHVM because of this limitation: https://github.com/facebook/hhvm/issues/1447

And the files are stored as PHP files to take advantage of PHP opcache which makes the read even faster than other caching strategies.
Here is the inspiration: https://blog.graphiq.com/500x-faster-caching-than-redis-memcache-apc-in-php-hhvm-dcd26e8447ad

If we go without opcache the code will be cleaner but it will be that much slower than the current approach. It will still be better than nothing though.

JanPetr:
My bad about the `eval()` and HHVM workaround.
Makes sense -> #millisecondsmatters

And how about checking in [assertCacheFileIsValid](https://github.com/algolia/algoliasearch-client-php/pull/195/files#diff-4e98072e8ef57737039dce5e1f0c1fb8R38) if HHVM && eval() is enabled?

rayrutjes:
You are right, we are missing some edge cases there that we should probably cover, nice catch @JanPetr ❤️

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,183 @@
...
+    private $ttl;
+
+    /**
+     * @param int|null    $ttl The time to live of the cache in seconds.
+     * @param string|null $file
+     *
+     */
+    public function __construct($ttl = null, $file = null)
+    {
+        $this->failingHostsCacheFile = null === $file ? $this->getDefaultCacheFile() : (string) $file;
```
JanPetr: I'd rewrite this line to be more explicit? Like that it's not really readable and I have to really focus to get what it does :)

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * @param array $data
+     */
+    private function writeFailingHostsCacheFile(array $data)
+    {
+        $json = json_encode($data);
+        if ($json === false) {
+            return;
+        }
+        @file_put_contents($this->failingHostsCacheFile, $json);
```
JanPetr: Is the `@` really needed? All conditions are checked when the class is created and by suppressing error here we might lose some valuable info for potential debugging.

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,183 @@
...
+     * @param string|null $file
+     *
+     */
+    public function __construct($ttl = null, $file = null)
+    {
+        $this->failingHostsCacheFile = null === $file ? $this->getDefaultCacheFile() : (string) $file;
+
+        $this->assertCacheFileIsValid($this->failingHostsCacheFile);
+
+        if (null === $ttl) {
```
JanPetr: Change to `$var === null` because of consistency.

src/AlgoliaSearch/ClientContext.php
```diff
@@ -129,6 +126,18 @@ public function __construct($applicationID, $apiKey, $hostsArray, $placesEnabled
         $this->algoliaUserToken = null;
         $this->rateLimitAPIKey = null;
         $this->headers = array();
+
+        if (null === $failingHostsCache) {
```
JanPetr: Change to `$var === null` because of consistency

src/AlgoliaSearch/InMemoryFailingHostsCache.php
```diff
@@ -0,0 +1,76 @@
...
+     * @var int
+     */
+    private $ttl;
+
+    /**
+     * @param int|null $ttl The time to live of the cache in seconds.
+     */
+    public function __construct($ttl = null)
+    {
+        if (null === $ttl) {
```
JanPetr: Change to `$var === null` because of consistency

src/AlgoliaSearch/InMemoryFailingHostsCache.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @param string $host
+     */
+    public function addFailingHost($host)
+    {
+        if (! in_array($host, self::$failingHosts)) {
+            // Keep a local cache of failed hosts in case the file based strategy doesn't work out.
+            self::$failingHosts[] = $host;
+
+            if (null === self::$timestamp) {
```
JanPetr: Change to `$var === null` because of consistency

src/AlgoliaSearch/InMemoryFailingHostsCache.php
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Get failing hosts from cache. This method should also handle cache invalidation if required.
+     * The TTL of the failed hosts cache should be 5mins.
+     *
+     * @return array
+     */
+    public function getFailingHosts()
+    {
+        if (null === self::$timestamp) {
```
JanPetr: Change to `$var === null` because of consistency

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,183 @@
...
+        @unlink($this->failingHostsCacheFile);
+    }
+
+    /**
+     * @return array
+     */
+    private function loadFailingHostsCacheFromDisk()
+    {
+        $json = @file_get_contents($this->failingHostsCacheFile);
+        if (false === $json) {
```
JanPetr: Change to `$var === false` because of consistency

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,183 @@
...
+    {
+        @unlink($this->failingHostsCacheFile);
+    }
+
+    /**
+     * @return array
+     */
+    private function loadFailingHostsCacheFromDisk()
+    {
+        $json = @file_get_contents($this->failingHostsCacheFile);
```
JanPetr: `file_get_contents` throws warning, not fatal - worth to remove `@` in order not to loose debugging information

src/AlgoliaSearch/FileFailingHostsCache.php
JanPetr: `file_put_contents` throws warning, not fatal - worth to remove `@`

rayrutjes:
Looks like this problem is related to https://github.com/sebastianbergmann/phpunit/issues/1684
I referenced the local PHPUnit binary. Let's see how that works out.

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * @return array
+     */
+    private function loadFailingHostsCacheFromDisk()
+    {
+        if (! file_exists($this->failingHostsCacheFile)) {
+            return array();
+        }
+
+        $json = file_get_contents($this->failingHostsCacheFile);
```
maxiloc: I think you need the `@` here

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,192 @@
...
+
+    /**
+     * @param $file
+     */
+    private function assertCacheFileIsValid($file)
+    {
+        $fileDirectory = dirname($file);
+
+        if (! is_writable($fileDirectory)) {
+            throw new \RuntimeException(sprintf('Cache file directory "%s" is not writable.', $fileDirectory));
```
maxiloc: I am not sure we want to throw non catched exceptions here

src/AlgoliaSearch/FileFailingHostsCache.php
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /**
+     * @param array $data
+     */
+    private function writeFailingHostsCacheFile(array $data)
+    {
+        $json = json_encode($data);
+        if ($json !== false) {
+            file_put_contents($this->failingHostsCacheFile, $json);
```
maxiloc: Same here we probably need the @

src/AlgoliaSearch/FileFailingHostsCache.php
rayrutjes: Yes we do, this strategy makes non sense if the infra does not allow it. This is exactly what Runtime exceptions are for ;)

If tomorrow we push this strategy by default, we will catch the exception and instantiate a fallback instead.

src/AlgoliaSearch/FileFailingHostsCache.php
rayrutjes: @JanPetr made a point on this subject. We should not obfuscate errors here. The errors here are only warnings so it makes sense to have them to be able to know what happens.

src/AlgoliaSearch/FileFailingHostsCache.php
rayrutjes: Same as before ;)

rayrutjes:
All good @JanPetr @maxiloc ?
We should then add a line to the docs.

JanPetr:
Good for me

rayrutjes:
Will merge and release on Monday.

PR #250, opened 2017-01-19
implement better disjunctive faceting

maxiloc:
You can add to algoliasearch.php:

```
$client = new AlgoliaSearch\Client('AJ0P3S7DWQ', '%%API_KEY%%');
$index = $client->initIndex('best_buy');

var_dump($index->search('test', ['facetFilters' => ["type:Wireless speakers", "brand:Beats by Dr. Dre", ["categories:Audio", 'categories:TV & Home Theater']], 'facets' => [], 'disjunctiveFacets' => ['type', 'brand', 'categories']]));
```

To run some test on the best buy instantsearch demo

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -412,7 +412,7 @@ class Index
      * Search inside the index.
      *
      * @param string $query the full text query
-     * @param mixed  $args  (optional) if set, contains an associative array with query parameters:
+     * @param mixed $args (optional) if set, contains an associative array with query parameters:
      *                      - page: (integer) Pagination parameter used to select the page to retrieve.
      *                      Page is zero-based and defaults to 0. Thus, to retrieve the 10th page you need to set page=9
      *                      - hitsPerPage: (integer) Pagination parameter used to select the number of hits per page.
@@ -488,8 +488,8 @@ class Index
      *                      duplicate value for the attributeForDistinct attribute are removed from results. For example,
      *                      if the chosen attribute is show_name and several hits have the same value for show_name, then
      *                      only the best one is kept and others are removed.
-     *
      * @return mixed
+     * @throws AlgoliaException
      */
     public function search($query, $args = null)
     {
@@ -497,6 +497,10 @@ class Index
             $args = array();
         }
         $args['query'] = $query;
+
+        if (isset($args['disjunctiveFacets'])) {
+            return $this->searchWithDisjunctiveFaceting($query, $args);
+        }
 
         return $this->client->request(
             $this->context,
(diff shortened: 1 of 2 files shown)


src/AlgoliaSearch/Index.php
```diff
@@ -409,10 +409,159 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
     }
 
     /**
+     * @param $queryParams
+     * @return array
+     */
+    private function getDisjunctiveQueries($queryParams)
+    {
+        $queriesParams = array();
+        $disjunctiveFacets = $queryParams['disjunctiveFacets']; // We now that there is at least one
```
rayrutjes: We should assert it in case we re-use the method at some point.
Currently we only assert that `count($queryParams['disjunctiveFacets'])>0`, not that it is an array

src/AlgoliaSearch/Index.php
```diff
@@ -409,10 +409,159 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
...
+    /**
+     * @param $query
+     * @param $args
+     * @return mixed
+     * @throws AlgoliaException
+     */
+    private function searchWithDisjunctiveFaceting($query, $args)
+    {
+        if (isset($args['filters'])) {
+            throw new AlgoliaException('You can not use disjunctive faceting and the filters parameter');
```
rayrutjes: We should throw \InvalidArgument exceptions in those cases and keep AlgoliaExceptions in the future for API related Exceptions.
Otherwise it makes it hard to know if we are dealing with runtime exceptions or logic exceptions.

maxiloc:
thanks @rayrutjes I updated the code, tell me what you think

src/AlgoliaSearch/Index.php
```diff
@@ -408,11 +408,145 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
...
+        return $filters;
+    }
+
+    /**
+     * @param $queryParams
+     * @return array
+     */
+    private function getDisjunctiveQueries($queryParams)
+    {
+        assert(
```
maxiloc: You suggested an assert if I remember well, but I changed the logic of the code and indeed it does not make sense anymore 😄

maxiloc: In fact it still makes sense.
All checks are done before without raising errors.
This assert is there only if you use the method alone for any reason

maxiloc: > We should assert it in case we re-use the method at some point.
Currently we only assert that count($queryParams['disjunctiveFacets'])>0, not that it is an array

rayrutjes: I suggested to assert that the values have the correct types, not to use the assert method ;)

rayrutjes: Sorry if I mislead you here :/

maxiloc: Np. I replaced the assert by condition + exception.

src/AlgoliaSearch/Index.php
```diff
@@ -408,11 +408,145 @@ public function deleteByQuery($query, $args = array(), $waitLastCall = true)
...
+    /**
+     * @param $queryParams
+     * @return array
+     */
+    private function getDisjunctiveQueries($queryParams)
+    {
+        if (isset($queryParams['disjunctiveFacets'])
+            && is_array($queryParams['disjunctiveFacets'])
+            && count($queryParams['disjunctiveFacets']) > 0) {
+            throw new \InvalidArgumentException('disjunctiveFacets needs to be an non empty array');
```
JanPetr: How about raising the exception directly from `search` method? With this implementation the exception is in `getDisjunctiveQueries` method, but the end user never called such method - he called `search`, so he might be confused.

src/AlgoliaSearch/Index.php
```diff
@@ -498,6 +632,12 @@ public function search($query, $args = null)
         }
         $args['query'] = $query;
 
+        $disjunctiveFacets = isset($args['disjunctiveFacets']) ? $args['disjunctiveFacets'] : array();
+
+        if (count($disjunctiveFacets) > 0) {
```
JanPetr: I'd shorten it to `if (isset($args['disjunctiveFacets') && count($args['disjunctiveFacets']) > 0)`

maxiloc: count("test") is equal to 1 and is not an array so it will do undefined things

src/AlgoliaSearch/Index.php
maxiloc: makes sense

src/AlgoliaSearch/Index.php
JanPetr: I probably don't understand. If somebody passes "test" in `$args['disjunctiveFacets']` your code won't catch it as well.

maxiloc: You were right, I got confused between two parts of the code.
Can you check the new code

JanPetr:
Nice @maxiloc! One more thing about the code structure - I usually structure methods in the file by order of usage. In here I'd go:
- search
- searchWithDisjunctiveFaceting
- getDisjunctiveQueries
- getAlgoliaFiltersArrayWithoutCurrentRefinement

It helps me in readability of the code. WDYT?

src/AlgoliaSearch/Index.php
```diff
@@ -498,6 +617,18 @@ public function search($query, $args = null)
         }
         $args['query'] = $query;
 
+        if (isset($queryParams['disjunctiveFacets'])) {
+            if (! is_array($queryParams['disjunctiveFacets']) || count($queryParams['disjunctiveFacets']) > 0) {
```
rayrutjes: This assertion should be part of `searchWithDisjunctiveFaceting`

src/AlgoliaSearch/Index.php
```diff
@@ -498,6 +617,18 @@ public function search($query, $args = null)
...
+                throw new \InvalidArgumentException('disjunctiveFacets needs to be an non empty array');
+            }
+
+            if (isset($args['filters'])) {
```
rayrutjes: This assertion should be part of `searchWithDisjunctiveFaceting`

src/AlgoliaSearch/Index.php
maxiloc: I think it's better here. I agree with honza, it's less clear as a user to get this error in another method
For me it's make more sense and it's more readable

rayrutjes: Well from a design perspective it's not optimal, because if you now decide to re-use `searchWithDisjunctiveFaceting`, data wouldn't be properly validated.

maxiloc: Make sense. I updated the code

JanPetr:
Good for me

PR #273, opened 2017-02-17
Implement attributesToRetrieve for getObjects method

### LINKED ISSUE #258: We are not able to specify attributesToRetrieve for getObjects

JanPetr:
Fixes #258

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -172,9 +172,8 @@ class Index
     /**
      * Get an object from this index.
      *
-     * @param $objectID             the unique identifier of the object to retrieve
-     * @param $attributesToRetrieve (optional) if set, contains the list of attributes to retrieve as a string
-     *                              separated by ","
+     * @param string    $objectID             the unique identifier of the object to retrieve
+     * @param string[]  $attributesToRetrieve (optional) if set, contains the list of attributes to retrieve
      *
      * @return mixed
      */
@@ -194,6 +193,10 @@ class Index
             );
         }
 
+        if (is_array($attributesToRetrieve)) {
+            $attributesToRetrieve = implode(',', $attributesToRetrieve);
+        }
+
         return $this->client->request(
             $this->context,
             'GET',
@@ -209,9 +212,8 @@ class Index
     /**
      * Get several objects from this index.
      *
-     * @param array $objectIDs the array of unique identifier of objects to retrieve
-     * @param $attributesToRetrieve (optional) if set, contains the list of attributes to retrieve as a string
-     *                              separated by ","
+     * @param array    $objectIDs            the array of unique identifier of objects to retrieve
+     * @param string[] $attributesToRetrieve (optional) if set, contains the list of attributes to retrieve
      *
      * @return mixed
      *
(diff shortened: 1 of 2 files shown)


src/AlgoliaSearch/Index.php
```diff
@@ -210,12 +210,14 @@ public function getObject($objectID, $attributesToRetrieve = null)
...
      * @param array $objectIDs the array of unique identifier of objects to retrieve
+     * @param $attributesToRetrieve (optional) if set, contains the list of attributes to retrieve as a string
+     *                              separated by ","
      *
      * @return mixed
      *
      * @throws \Exception
      */
-    public function getObjects($objectIDs)
+    public function getObjects($objectIDs, $attributesToRetrieve = null)
```
rayrutjes: I think we should ask for an array of strings like we do for the other methods.
The documentation is bad for some of the other methods, we say we expect comma separated values where we are actually accepting arrays.

JanPetr: I wanted to be consistent with [`getObject` method](https://github.com/algolia/algoliasearch-client-php/blob/master/src/AlgoliaSearch/Index.php#L181) where we take comma-separated string.
Definitely not something I'd approve and I'd definitely rather accept `array`, but for me consistency has higher priority right now.

rayrutjes: Arg, I just tested and it doesn't work with an array in the 'getObject' method.
I would:
- Make sure we support arrays in both `getObject` and `getObjects`
- Change docbloc and documentation to array
- Keep backward compat

JanPetr: Good idea, can you review @rayrutjes?

src/AlgoliaSearch/Index.php
```diff
@@ -210,12 +214,14 @@ public function getObject($objectID, $attributesToRetrieve = null)
      * Get several objects from this index.
      *
      * @param array $objectIDs the array of unique identifier of objects to retrieve
+     * @param array $attributesToRetrieve (optional) if set, contains the list of attributes to retrieve as a string
```
rayrutjes: I would change the docblock here to 'array of strings' en not mention the 'separated by ,' anymore

src/AlgoliaSearch/Index.php
```diff
@@ -224,6 +230,15 @@ public function getObjects($objectIDs)
         $requests = array();
         foreach ($objectIDs as $object) {
             $req = array('indexName' => $this->indexName, 'objectID' => $object);
+
+            if ($attributesToRetrieve) {
```
rayrutjes: This will fail if empty array is provided.

JanPetr: That's correct behavior. Empty array = all attributes are fetched

rayrutjes: It is not the current behaviour. Empty string returns only objectIDs.

JanPetr: It's the API behavior for `getObject` endpoint. For `getObjects` API returns all attributes on empty string/array.

rayrutjes:
LGTM!

PR #275, opened 2017-02-23
rename userKey to apiKey

maxiloc:


JanPetr:
[not critical] How about writing all "API"s in capitals as it's acronym?

maxiloc:
> [not critical] How about writing all "API"s in capitals as it's acronym?

I thought about it but it's less readable and less searchable because we can not uncamelize it.

JanPetr:
@maxiloc Makes sense. But still at least in comments it'd make it more readable.

maxiloc:
that we can do 😄

src/AlgoliaSearch/Client.php
```diff
@@ -408,13 +408,13 @@ public function initIndex($indexName)
...
     /**
-     * List all existing user keys with their associated ACLs.
+     * List all existing API keys with their associated ACLs.
      *
      * @return mixed
      *
      * @throws AlgoliaException
      */
-    public function listUserKeys()
+    public function listApiKeys()
```
rayrutjes: If we are into consistency, I'd rather go with `getApiKeys` unless this is already named otherwise in the other clients.

src/AlgoliaSearch/Index.php
```diff
@@ -990,13 +990,13 @@ public function setSettings($settings, $forwardToReplicas = false)
...
     /**
-     * List all existing user keys associated to this index with their associated ACLs.
+     * List all existing API keys associated to this index with their associated ACLs.
      *
      * @return mixed
      *
      * @throws AlgoliaException
      */
-    public function listUserKeys()
+    public function listApiKeys()
```
rayrutjes: Same here

src/AlgoliaSearch/Client.php
maxiloc: I would prefer to keep the distinction between get and list.
For me list implies that you don't have to specifies which key to get, it will list them all.
getKeys does not imply that. And since getKey takes a key as a parameters, I would expect getKeys to take an array of keys as parameter which is not the case.

rayrutjes: I don't agree with your arguments but I do agree to keep it like it is given it is implemented the same way in other clients. 
Go for it!

PR #278, opened 2017-03-27
test(keep-alive): update the connection timeout

rayrutjes:
The default connection timeout is now 1s but the keep-alive test still assumed it was 2.
Locally my tests fail because of this.

BTW I'm not seing how this tests actually tests keep alive, any insight @maxiloc ?

### DIFF
--- a/tests/AlgoliaSearch/Tests/FunctionTest.php
+++ b/tests/AlgoliaSearch/Tests/FunctionTest.php
@@ -33,16 +33,16 @@ class FunctionTest extends AlgoliaSearchTestCase
             $this->markTestSkipped('Irrevelant on travis');
         }
         $this->client = new Client(getenv('ALGOLIA_APPLICATION_ID'), getenv('ALGOLIA_API_KEY'));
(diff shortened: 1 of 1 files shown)


maxiloc:
It test that the first query is longer than the average of the 10 next one.
The test is definitely not perfect. but I don't see a better way to do this

rayrutjes:
I see it now. So in the end why don't we do the following: 
`$this->assertLowerThan($timeAfter, $timeInit);`

If that is the intention, then I'll gladly refresh a bit the naming in the test to make it more explicit.

Regarding the strategy I think it's good.

maxiloc:
yep that would be better

rayrutjes:
Ok, I will do that 👍

rayrutjes:
This is planned @JanPetr ;)
I added you by mistake on this and wanted to re-write everything because I 100% agree with your statement, this test is un-readable ;)

rayrutjes:
I used your naming @JanPetr 👍

PR #282, opened 2017-04-19
Add section with informations about testing

roukmoute:
| Q             | A
| ------------- | ---
| Doc?         | yes

### Description

This PR adds new section about run tests correctly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,47 @@
+## Contributing
+
+Hi there! We're thrilled that you'd like to contribute to this project.
+Your help is essential for keeping it great.
+
+## Submitting a pull request
+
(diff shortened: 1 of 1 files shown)


rayrutjes:
Thanks @roukmoute , this PR makes a lot of sense.

For now the README.md is generated by an internal tool, and I don't think we want this kind of information in it.

That being said, I'd gladly have a CONTRIBUTING.md file at the root of the directory containing this kind of information.

WDYT?

rayrutjes:
Could be a good place to also mention the fact that builds are expected to fail on PR, because we have no way of sharing TRAVIS private env variables.

roukmoute:
@rayrutjes I've updated my commit

Issue #283, opened 2017-04-20
Add differents AlgoliaException code


roukmoute:
It would be nice to have differents exceptions code to treat them differently.

For example with a RabbitMQ, if we receive an host unreacheable error, we would like to send a retry exchange.

PR #284, opened 2017-04-20
Exception code

roukmoute:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests?        | yes
| Doc?          | no
| Fixed tickets | https://github.com/algolia/algoliasearch-client-php/issues/283

### Description

This PR catches an Exception code to distinct different type of AlgoliaException..

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -712,7 +712,6 @@ class Client
      */
     public static function generateSecuredApiKey($privateApiKey, $query, $userToken = null)
     {
-        $urlEncodedQuery = '';
         if (is_array($query)) {
             $queryParameters = array();
             if (array_keys($query) !== array_keys(array_keys($query))) {
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -819,7 +819,7 @@ class Client
                 }
             }
         }
-        throw new AlgoliaException('Hosts unreachable: '.implode(',', $exceptions));
+        throw new AlgoliaConnectionException('Hosts unreachable: '.implode(',', $exceptions));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


rayrutjes:
Hey, thanks @roukmoute ,
Can you add a test to cover that new code?

Also I'm unsure what the current error code is, and if users rely on that code already in which case this PR would introduce a BC somehow.

Happy to discuss it ;)

roukmoute:
I don't think how it could introduce BC, because actually all code exceptions are at `0`.
So no one can use this code.

Now on the contrary it could help to know what kind of exception it is.

My [issue](#283) is a real concrete example that I have on a project.

rayrutjes:
Thanks for the precisions @roukmoute , let's move forward with this then, I see no reason to not ge a proper exception code here.

I feel like there are other places where we could set the correct error codes like here: https://github.com/roukmoute/algoliasearch-client-php/blob/3f1a3b823e6d0d33373a807b393a5843aab626a1/src/AlgoliaSearch/Client.php#L992

If you can just add a unit test for the current code you have, then eventually address other error codes then I can take it over.

Best.

roukmoute:
Here it is my update :)

roukmoute:
ping @rayrutjes @maxiloc

rayrutjes:
Thanks for the ping @roukmoute , we will take a look at it.
Please bear with us.

src/AlgoliaSearch/Client.php
```diff
@@ -989,7 +989,7 @@ public function doRequest(
         curl_close($curlHandle);
 
         if (intval($http_status / 100) == 4) {
-            throw new AlgoliaException(isset($answer['message']) ? $answer['message'] : $http_status.' error');
+            throw new AlgoliaException(isset($answer['message']) ? $answer['message'] : $http_status.' error', $http_status);
         } elseif (intval($http_status / 100) != 2) {
             throw new \Exception($http_status.': '.$response);
```
JanPetr: Shouldn't the code be added to this exception as well?

roukmoute: Done @JanPetr

roukmoute:
Done @JanPetr

roukmoute:
Ping @rayrutjes @maxiloc @JanPetr

rayrutjes:
Sorry about the delay, please check #298 .

roukmoute:
Why have you closed my P.R.?

rayrutjes:
Hey @roukmoute ,
Your commits have been pushed as part of #298, you will get credits there.

Because we are actually using secured tokens from Travis, we need to push code ourselves for the tests to run.

Issue #285, opened 2017-04-26
\AlgoliaSearch\Index::__call should throw an exception on unhandled methods


longwave:
We just had to track down a silent failure bug where we were calling an invalid method on an index object, but not seeing any errors. This is because the `__call` implementation simply returns NULL if the method name is not recognised; only `browse` is actually handled by this method.

It would be good if an exception was thrown here to avoid others running into the same problem.

rayrutjes:
Thanks for reporting this @longwave .

I just submitted a PR and will wait for the team mates to approve it ;)

Cheers!

longwave:
Honestly amazed at the speed of response here, thank you!

PR #286, opened 2017-04-26
Fix/285

rayrutjes:
Resolves: #285


JanPetr:
Wow, cannot believe we missed such an issue.
The fix introduces a BC break which will break already "broken" implementations. But I cannot see any "non-BC-break" way how to fix it.

rayrutjes:
Technically I agree with you @JanPetr , but it is more of a bug fix in this case.
If people are relying on bugs, then we cannot fix them.

For this one, I think we are good, and for eventual breaking apps, the fix will be more than obvious.

tests/AlgoliaSearch/Tests/IndexTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+class IndexTest extends AlgoliaSearchTestCase
+{
+    /**
+     * @expectedException \BadMethodCallException
+     * @expectedExceptionMessage No method named unknown was found.
+     */
+    public function testShouldThrowAnExceptionIfUnknownMethodIsCalled()
+    {
+        $clientContextMock = $this->getMockBuilder('AlgoliaSearch\ClientContext')->disableOriginalConstructor()->getMock();
```
JanPetr: What is the reason behind using mock instead of real Client(Context) with passed env API credentials? 
Would it make sense to do it as unit test in a different test suite?

rayrutjes: Here I mock the dependencies to test the Index class in isolation.
This is a unit test.

JanPetr: I know it's a unit test. I was wondering if it makes sense to write a unit test and if it does if it's not better to write it in a different namespace / folder to create separate (unit and integration) test suites.

rayrutjes: Yes we could separate integration from unit tests.
Let's do it in another PR though because we would need to separate more things than only this test.

PR #294, opened 2017-06-21
feat(query-rules): add query-rules support

rayrutjes:
I expect the tests to fail on travis given the feature is not deployed on our testing APP ID.

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -1625,6 +1625,10 @@ class Index
      */
     public function saveRule($objectID, $content, $forwardToReplicas = false)
     {
+        if (!isset($content['objectID'])) {
+            $content['objectID'] = $objectID;
+        }
+
         return $this->client->request(
             $this->context,
             'PUT',
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -1506,6 +1506,138 @@ class Index
     }
 
     /**
+     * @param $params
+     *
+     * @return mixed
+     *
+     * @throws AlgoliaException
+     */
+    public function searchRules(array $params = array())
+    {
+        return $this->client->request(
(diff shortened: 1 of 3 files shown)


clement-leprovost:
@rayrutjes The API is deployed everywhere: we can merge it.

rayrutjes:
Waiting for the tests to pass and for an additional review.

rayrutjes:
@julienbourdeau @maxiloc @JanPetr 
Can any of you give this a review?

src/AlgoliaSearch/Index.php
```diff
@@ -1506,6 +1506,142 @@ public function searchFacet($facetName, $facetQuery, $query = array())
...
+    public function saveRule($objectID, $content, $forwardToReplicas = false)
+    {
+        if (!isset($content['objectID'])) {
+            $content['objectID'] = $objectID;
+        }
+
+        return $this->client->request(
+            $this->context,
+            'PUT',
+            '/1/indexes/'.$this->urlIndexName.'/rules/'.urlencode($objectID).'?forwardToReplicas='.($forwardToReplicas ? 'true' : 'false'),
```
julienbourdeau: I feel like if `$content['objectID']` is different from `$objectID` we might be updating the wrong rule. I'm not sure I understand why we need to pass an $objectID.

clement-leprovost: The REST API requires the `objectID` to be part of the URL *and* also part of the payload. I know it's not very RESTful, but it's the way it works at the moment. Don't worry, the API will reject the request if the two object IDs mismatch.

clement-leprovost:
@julienbourdeau `saveRule` is not a partial update, it's a full update. It's called "save" because it can be either a create or an update (PUT method).

Note: The API for rules is essentially the same as that for [synonyms](https://www.algolia.com/doc/api-client/php/synonyms/).

PR #296, opened 2017-07-05
Add $requestHeaders parameters to necessary methods

julienbourdeau:
See internal conversation for context 
https://algolia.slack.com/archives/C3742QTHT/p1499242553433222

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -83,7 +83,7 @@ class Index
      *
      * @throws \Exception
      */
-    public function batchObjects($objects, $objectIDKey = 'objectID', $objectActionKey = 'objectAction')
+    public function batchObjects($objects, $objectIDKey = 'objectID', $objectActionKey = 'objectAction', $requestHeaders = array())
     {
         $requests = array();
         $allowedActions = array(
@@ -114,7 +114,7 @@ class Index
             $requests[] = $req;
         }
 
-        return $this->batch(array('requests' => $requests));
+        return $this->batch(array('requests' => $requests), $requestHeaders);
     }
 
     /**
@@ -127,7 +127,7 @@ class Index
      *
      * @return mixed
      */
-    public function addObject($content, $objectID = null)
+    public function addObject($content, $objectID = null, $requestHeaders = array())
     {
         if ($objectID === null) {
             return $this->client->request(
@@ -138,7 +138,8 @@ class Index
                 $content,
                 $this->context->writeHostsArray,
                 $this->context->connectTimeout,
-                $this->context->readTimeout
+                $this->context->readTimeout,
+                $requestHeaders
             );
         }
 
@@ -150,7 +151,8 @@ class Index
             $content,
             $this->context->writeHostsArray,
             $this->context->connectTimeout,
-            $this->context->readTimeout
+            $this->context->readTimeout,
+            $requestHeaders
         );
     }
 
@@ -162,11 +164,11 @@ class Index
      *
      * @return mixed
      */
-    public function addObjects($objects, $objectIDKey = 'objectID')
+    public function addObjects($objects, $objectIDKey = 'objectID', $requestHeaders = array())
     {
         $requests = $this->buildBatch('addObject', $objects, true, $objectIDKey);
 
-        return $this->batch($requests);
+        return $this->batch($requests, $requestHeaders);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


maxiloc:
👍 but, like you, not in favor of merging it, it makes very little sense to have it for all methods

rayrutjes:
This makes a lot of sense in languages that share the client between different requests.
In PHP though, I don't think we need this.

maxiloc:
We need to add indexing methods

rayrutjes:
alright will do that.

rayrutjes:
This PR doesn't feel right.
Did we think of adding the info to the ClientContext @maxiloc ?

rayrutjes:
We already have a way to do this: 

```
$index = $client->initIndex('your_index_name');
$client->setExtraHeader('X-Algolia-User-ID', 'user123');
$res = $index->search('query string');
```

I don't think we need more for now.

maxiloc:
Looks good to me 👍 . @julienbourdeau Maybe double check it as well

src/AlgoliaSearch/Index.php
```diff
@@ -682,7 +682,7 @@ private function getAlgoliaFiltersArrayWithoutCurrentRefinement($filters, $needl
      *
      * @return mixed
      */
-    public function searchForFacetValues($facetName, $facetQuery, $query = array(), $requestHeaders)
+    public function searchForFacetValues($facetName, $facetQuery, $query = array(), $requestHeaders = array())
```
julienbourdeau: Good catch, thanks!

julienbourdeau:
Looks good to me to 🚀

alcaeus:
For what it's worth, this is a BC break that causes the AlgoliaSearchBundle for Symfony to fail:
```
  Warning: Declaration of Algolia\AlgoliaSearchBundle\Debug\DebugClient::doRe  
  quest($context, $method, $host, $path, $params, $data, $connectTimeout, $re  
  adTimeout) should be compatible with AlgoliaSearch\Client::doRequest($conte  
  xt, $method, $host, $path, $params, $data, $connectTimeout, $readTimeout, $  
  requestHeaders = Array)                                                      
```

Adding a parameter to a public method in a non-final class (even with default value) is a BC break that should not be done in a minor release.

rayrutjes:
arf yes @alcaeus we will act fast.

alcaeus:
Wow. I'm **extremely** disappointed. Let me break it down, maybe you understand my frustration:

> All classes should be considered final

"considered final" is not a thing. The class is either `final` or it isn't. If it is, do as you please. If it isn't, respect BC - simple as that. If you want to have the classes final, deprecate extending them in 1.24.0 and make them `final` in 2.0.

> and should be inherited by composition.

I can't inherit by composition if you don't provide an interface. The `Client` class needs to be extended in order for functionality to be added. Once you provide an interface, I'll add my changes by composing the class. It doesn't change anything though, as changing the method signature in an interface is only allowed in major releases as well, so it would've been a BC break either way.

> but we have to take a decision between having users with outdated version because they locked their version to 1, or introducing this breaking change.

"We have to either get people to update our software or force them to update theirs by breaking it". This is exactly why Semantic Versioning exists: I want to update from 1.21.0 -> 1.23.0 with confidence and without being afraid of BC breaks. Guess what: I'll be on 1.22.0 for a while because I don't appreciate you breaking my code with minor releases. This is exactly how you end up with users with outdated versions.

If you want to add a new parameter without breaking BC, take a look at how Symfony solves it: https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/DependencyInjection/Compiler/PassConfig.php#L121..L134
TL;DR:
- Add param in docs only
- Read param using `func_num_args` and `func_get_args`
- Throw deprecation notice if method was overridden in an inherited class
- Fix properly in next major version

That's how you make changes like this one, not by breaking any custom implementation.

> Could you update the signature of the method where needed in your code or the library you are using

No. For now, I've locked the client dependency to `<1.23.0` - that keeps the code working and doesn't require me to fix your mistakes on our end. Eventually, maybe.

Just in case you're interested in the code I need to update: the `DebugClient` mentioned in the error message above is part of a contribution made by our team to your Symfony bundle (see https://github.com/algolia/AlgoliaSearchBundle/pull/57). Maybe having it merged would've prevented this kind of BC break in a minor release, maybe it wouldn't have.

rayrutjes:
@alcaeus I appreciate the candid feedback here. We will see if using `func_get_args` could make sense here. I'm afraid it will make the code we have really unreadable but worth considering in order to avoid the breaking changes.

PR #298, opened 2017-08-02
Connection exception

rayrutjes:


rayrutjes:
Let's move forward with #284 .
I'd love a second pair of eyes here.

roukmoute:
Seriously!?

I created a P.R. several months ago now and you close MY P.R. that helps YOU?

But you did not quite understand the concept of open source!

And your excuse is [Travis](https://github.com/algolia/algoliasearch-client-php/pull/284#issuecomment-319791681)?

This is the first time I see that.  
Just keep off travis and take another product.

But it would still be the least of the things after this terrible delay while we pay for your service!

rayrutjes:
@roukmoute ,

I'm very sorry about the delay of handling your contribution, and we really appreciate you took the time to do the PR.

Your commits have been left as is, and you are still the author of the code and will get the credits when merged.

We are unfortunately unable to run tests automatically on community PR's, because it would expose the secret credentials.
We looked a long time for a solution, but so far, we have found none.
If you have idea on how to solve it, we'd gladly move forward.

The issue is not really Travis. We would have the same issue with other providers.
Given we use secret credentials, we can not run code if it has not been previously approved.

PR #326, opened 2017-10-10
Introduce SynonymsBrowser class

julienbourdeau:
This class allows you to easily browse (or export) all your synonyms, just like you browse the records of your index.

I would like feedback on how the array of a single synonym should be formatted. Currently, it's as follow:

```
Array
(
    [type] => altCorrection1
    [word] => Avenue
    [corrections] => Array
        (
            [0] => Av
            [1] => Ave
        )

    [objectID] => avenue
)
```

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -1509,9 +1509,9 @@ class Index
         );
     }
 
-    public function exportSynonyms(array $synonymType = array(), $batchSize = 500)
+    public function exportSynonyms($batchSize = 500)
     {
-        return new SynonymBrowser($this, $synonymType, 0, $batchSize);
+        return new SynonymBrowser($this, $batchSize);
     }
 
     /**
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -1509,7 +1509,7 @@ class Index
         );
     }
 
-    public function exportSynonyms($batchSize = 500)
+    public function initSynonymBrowser($batchSize = 1000)
     {
         return new SynonymBrowser($this, $batchSize);
     }
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -1509,6 +1509,11 @@ class Index
         );
     }
 
+    public function exportSynonyms(array $synonymType = array(), $batchSize = 500)
+    {
+        return new SynonymBrowser($this, $synonymType, 0, $batchSize);
+    }
+
     /**
      * @deprecated Please use searchForFacetValues instead
      * @param $facetName
--- a/src/AlgoliaSearch/SynonymBrowser.php
+++ b/src/AlgoliaSearch/SynonymBrowser.php
@@ -11,19 +11,14 @@ class SynonymBrowser implements \Iterator
     private $index;
 
     /**
-     * @var int Zero-based page number
-     */
-    private $page = 0;
-
-    /**
      * @var int Number of results to return from each call to Algolia
      */
     private $hitsPerPage;
 
     /**
-     * @var int Position in the current paginated result batch
+     * @var int
      */
-    private $position = 0;
+    private $key = 0;
 
     /**
      * @var array Response from the last Algolia API call,
@@ -49,8 +44,9 @@ class SynonymBrowser implements \Iterator
     public function current()
     {
         $this->ensureResponseExists();
+        $hit = $this->response['hits'][$this->getHitIndexForCurrentPage()];
 
-        return $this->formatHit($this->response['hits'][$this->position]);
+        return $this->formatHit($hit);
     }
 
     /**
@@ -59,30 +55,21 @@ class SynonymBrowser implements \Iterator
      */
     public function next()
     {
-        $this->ensureResponseExists();
-
-        $this->position++;
-
-        if ($this->valid()) {
-            return;
-        }
-
-        // If the end of the batch is reached but there are
-        // more results in Algolia, we get the next page.
-        if (count($this->response['hits']) >= $this->hitsPerPage) {
-            $this->page++;
-            $this->position = 0;
-            $this->fetchCurrentPageResults();
+        $previousPage = $this->getCurrentPage();
+        $this->key++;
+        if($this->getCurrentPage() !== $previousPage) {
+            // Discard the response if the page has changed.
+            $this->response = null;
         }
     }
 
     /**
      * Return the key of the current element
-     * @return mixed scalar on success, or null on failure.
+     * @return int
      */
     public function key()
     {
-        return $this->page * $this->hitsPerPage + $this->position;
+        return $this->key;
     }
 
     /**
@@ -97,7 +84,7 @@ class SynonymBrowser implements \Iterator
     {
         $this->ensureResponseExists();
 
-        return isset($this->response['hits'][$this->position]);
+        return isset($this->response['hits'][$this->getHitIndexForCurrentPage()]);
     }
 
     /**
@@ -106,21 +93,8 @@ class SynonymBrowser implements \Iterator
      */
     public function rewind()
     {
-        $this->page = 0;
-        $this->position = 0;
+        $this->key = 0;
         $this->response = null;
-
-        $this->fetchCurrentPageResults();
-    }
-
-    /**
-     * Call Algolia' API to get new result batch
-     *
-     * @param $page Page number to call
-     */
-    private function fetchCurrentPageResults()
-    {
-        $this->response = $this->index->searchSynonyms('', array(), $this->page, $this->hitsPerPage);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/AlgoliaSearch/SynonymBrowser.php
```diff
@@ -0,0 +1,158 @@
...
+    private $response;
+
+    /**
+     * SynonymBrowser constructor.
+     * @param Index $index
+     * @param array $synonymType
+     * @param int $page
+     * @param int $hitsPerPage
+     */
+    public function __construct($index, array $synonymType = array(), $page = 0, $hitsPerPage = 500)
```
rayrutjes: I think we need to conform with other API Clients: https://github.com/algolia/algoliasearch-client-python/pull/297/files

rayrutjes: The signature should be `Index $index, $hitsPerPage`.

rayrutjes:
We also need to implement the same for "rules". Maybe we can first iterate on this one, and then copy the logic.

src/AlgoliaSearch/SynonymBrowser.php
julienbourdeau: Indeed! I'll edit 👌

src/AlgoliaSearch/SynonymBrowser.php
```diff
@@ -0,0 +1,158 @@
...
+{
+    /**
+     * @var Index
+     */
+    private $index;
+
+    /**
+     * @var array Used to filter the type of synonyms you want to export
+     */
+    private $synonymType;
```
clement-leprovost: I don't think this is required. A basic export function to export all synonyms & placeholders would be fine IMHO. (But of course, it doesn't harm if you think it's useful.)

julienbourdeau: I can't really thing a good use case where you need to export only one type of synonyms to be honest. I added it because it was in the function signature of `searchSynonyms` and I thought it would be nice if they are similar.
But I'd rather stay closer to the other API clients, it makes more sense.

src/AlgoliaSearch/SynonymBrowser.php
```diff
@@ -44,25 +39,22 @@ class SynonymBrowser implements \Iterator
     /**
      * SynonymBrowser constructor.
      * @param Index $index
-     * @param array $synonymType
-     * @param int $page
      * @param int $hitsPerPage
      */
-    public function __construct($index, array $synonymType = array(), $page = 0, $hitsPerPage = 500)
+    public function __construct($index, $hitsPerPage = 500)
```
rayrutjes: Can we type hint the `$index`?

rayrutjes: We should also use `1000` by default as in the python implementation.

src/AlgoliaSearch/SynonymBrowser.php
```diff
@@ -0,0 +1,150 @@
...
+        // No more results and it was the last page
+        if (count($this->response['hits']) < $this->hitsPerPage) {
+            return false;
+        }
+
+        // No more results but still results in Algolia
+        // We call Algolia and loop back to this function
+        $this->doQuery($this->page + 1);
+
+        return $this->valid();
```
rayrutjes: The `valid` method should not have side effects and it should be idempotent.

src/AlgoliaSearch/SynonymBrowser.php
```diff
@@ -0,0 +1,150 @@
...
+     * is not valid, we call Algolia' API to load more results
+     * until it's the last page.
+     *
+     * @return boolean The return value will be casted to boolean and then evaluated.
+     * Returns true on success or false on failure.
+     */
+    public function valid()
+    {
+        if (isset($this->response['hits'][$this->position])) {
+            $this->hit = $this->formatHit($this->response['hits'][$this->position]);
```
rayrutjes: `formatHit` is only used once. I would inline the single delete statement it contains.

src/AlgoliaSearch/Index.php
```diff
@@ -1509,6 +1509,11 @@ public function saveSynonym($objectID, $content, $forwardToReplicas = false)
         );
     }
 
+    public function exportSynonyms($batchSize = 500)
```
rayrutjes: I think `initSynonymBrowser` or `newSynonymBrowser` might be a better name.

src/AlgoliaSearch/SynonymBrowser.php
```diff
@@ -0,0 +1,150 @@
...
+     */
+    public function __construct($index, $hitsPerPage = 500)
+    {
+        $this->index = $index;
+        $this->hitsPerPage = $hitsPerPage;
+
+        $this->page = 0;
+        $this->position = 0;
+
+        $this->doQuery($page);
```
rayrutjes: Calls to Algolia should only be done when required. We could lazy load results.

tests/AlgoliaSearch/Tests/SynonymsExportTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->client = new Client(getenv('ALGOLIA_APPLICATION_ID'), getenv('ALGOLIA_API_KEY'));
+        $this->index = $this->client->initIndex($this->safe_name('test-synonym-export-php'));
+
+        try {
+            $this->index->clearIndex();
+        } catch (AlgoliaException $e) {
+            // not fatal
+        }
+
+        $res = $this->index->addObject(array('name' => '589 Howard St., San Francisco'));
```
rayrutjes: Is this to create the index?
I presume it is not possible to add synonyms to a non existing index?
Doing a `setSettings([])` might be a bit more readable.

tests/AlgoliaSearch/Tests/SynonymsExportTest.php
```diff
@@ -0,0 +1,81 @@
...
+        $this->index->waitTask($res['taskID'], 0.1);
```
rayrutjes: We don't need to wait for this one, given tasks will be processed in order and we are waiting for the batched synonyms right after.

tests/AlgoliaSearch/Tests/SynonymsExportTest.php
```diff
@@ -0,0 +1,81 @@
...
+    /** @var Index */
+    private $index;
+
+    protected function setUp()
+    {
+        $this->client = new Client(getenv('ALGOLIA_APPLICATION_ID'), getenv('ALGOLIA_API_KEY'));
+        $this->index = $this->client->initIndex($this->safe_name('test-synonym-export-php'));
+
+        try {
+            $this->index->clearIndex();
```
rayrutjes: We need to clear synonyms here.

src/AlgoliaSearch/SynonymBrowser.php
julienbourdeau: Yes good idea. I guess I was too inspired by the IndexBrowser. I'll remove it 👌

rayrutjes: Let's raise the bar! 🚀

tests/AlgoliaSearch/Tests/SynonymsExportTest.php
julienbourdeau: Yes, I'll remove it 👌

tests/AlgoliaSearch/Tests/SynonymsExportTest.php
julienbourdeau: I got `No content in PUT request` error with empty settings. I will make it more obvious that this call is to create the index.

tests/AlgoliaSearch/Tests/SynonymsExportTest.php
julienbourdeau: Oops ^^

src/AlgoliaSearch/SynonymBrowser.php
julienbourdeau: But we need to call algolia go load new results, I cannot see a better way to do it.

src/AlgoliaSearch/SynonymBrowser.php
julienbourdeau: In order to skip function call? I find it more readable with a function ¯\_(ツ)_/¯

src/AlgoliaSearch/SynonymBrowser.php
rayrutjes: The `rewind` method is always called first and should be used to load the first response.
Then when doing `next`, we could check if we reached the end of the current response and load next one.

src/AlgoliaSearch/SynonymBrowser.php
rayrutjes: I'm fine with that. It's internal anyway so we can iterate on that if needed 👍

src/AlgoliaSearch/SynonymBrowser.php
julienbourdeau: I see what you mean, I like the idea.

julienbourdeau: Well actually `doQuery` should be called in `rewind` anyway, otherwise you won't access the correct value 😱

julienbourdeau:
I'm adding a test to import all exported synonyms and it fails. I believe the export method should format all item so they can be imported easily into Algolia. I'll look into it in details.

cc @clement-leprovost

src/AlgoliaSearch/Index.php
```diff
@@ -1509,6 +1509,11 @@ public function saveSynonym($objectID, $content, $forwardToReplicas = false)
         );
     }
 
+    public function initSynonymBrowser($batchSize = 500)
```
rayrutjes: Should probably be 1000 here as well

julienbourdeau:
I think it's a good idea to calculate page, so we get rid of $position and $page attribute and only rely on $key. 

I think we're back in a state where `valid` triggers the call to populate the `$response`, the method not idempotent anymore

rayrutjes:
I was under the impression that in the previous implementation there were some side effects, but at a second look it actually was fine regarding.

With the current implementation though, we only call Algolia when strictly required.

Also, every methods work and abstract the fact that there is an API call happening behind the scenes.

PR #328, opened 2017-10-16
Introduce RuleIterator

julienbourdeau:
Most of the Iterator code was extracted to `AlgoliaIterator` and both synonyms and rules rely extend this class.

Rules are limited to 500 by default because the API will currently reject 1000. The python API clients will update to 500 too.


julienbourdeau:
OMG I forgot about this file.

I used 500 because 1000 is out of range for the Rule search endpoint. 999 works but not 1000, I thought 500 would be better, Clement will change it in the Python client PR too.

rayrutjes:
Awesome, then let's just make sure we add the 3 classes to the classmap ;)

rayrutjes:
All good here, go for the merge!

PR #329, opened 2017-10-16
Upgrade config for php-cs-fixer v2

julienbourdeau:


rayrutjes:
Can we add the command as a dev dependency as well?
Maybe also add it as a composer script could help us running it more often.

julienbourdeau:
I tried but as far as I understood, we can't because we require `php >= 5.3` and it seems incompatible. I'll see if I find a way.
At least I wanted to be able to use it locally (my test failed previously because I used the array short syntax).

rayrutjes:
php-cs-fixer is super confusing, and the fact that we still need to support 5.3 definitely makes it trickier :/

.php_cs
```diff
@@ -10,6 +10,10 @@ return PhpCsFixer\Config::create()
     ->setRules([
         '@Symfony' => true,
         'array_syntax' => ['syntax' => 'long'],
+        'trailing_comma_in_multiline_array' => false,
```
julienbourdeau: I think this was on before. Why not leaving the Symfony default?

rayrutjes: No it was off. When I turn it on, it adds a trailing coma everywhere.
I personally prefer it on though. I was about to ask you your feedback on the PR.

rayrutjes:
Let me know if there are rules you'd like to amend @julienbourdeau @JanPetr

julienbourdeau:
Rebased and removed commit merge. I'll fix the master branch after it's merged (to avoid conflicts while this PR is pending).

julienbourdeau:
Because the v2 is around the corner, I'd rather avoid having this kind of big change on the v1 now. I'll close it, if we really need to fix the codebase, we can do it but I think we're fine. Thanks Raymond for the help.

PR #331, opened 2017-10-25
feat(MCM): add new MCM endpoints

Nagriar:


src/AlgoliaSearch/Client.php
```diff
@@ -404,6 +404,217 @@ public function getLogs($offset = 0, $length = 10, $type = 'all', $requestHeader
...
+        );
+    }
+
+    /**
+     * Remove a userID from the mapping
+     * @return an object containing a "deletedAt" attribute
+     *
+     * @throws AlgoliaException
+     */
+    public function RemoveUserID($userID, $requestHeaders = array())
```
rayrutjes: "R" should be lowercased

src/AlgoliaSearch/Client.php
```diff
@@ -404,6 +404,217 @@ public function getLogs($offset = 0, $length = 10, $type = 'all', $requestHeader
...
+     *     ),
+     *     "nbHits" => 0
+     *     "page" => 0,
+     *     "hitsPerPage" => 20
+     * ).
+     *
+     * @return mixed
+     * @throws AlgoliaException
+     */
+    public function searchUserIDs($query, $clusterName = null, $page = null, $hitsPerPage = null, $requestHeaders = array())
```
rayrutjes: We should probably merge the 4 first params together.

Nagriar: I used the same format than `searchSynonyms` https://github.com/algolia/algoliasearch-client-php/blob/master/src/AlgoliaSearch/Index.php#L1397

src/AlgoliaSearch/Client.php
```diff
@@ -404,6 +404,215 @@ public function getLogs($offset = 0, $length = 10, $type = 'all', $requestHeader
...
+            $params['hitsPerPage'] = $hitsPerPage;
+        }
+
+        return $this->request(
+            $this->context,
+            'POST',
+            '/1/clusters/mapping/search',
+            null,
+            $params,
+            $this->context->readHostsArray,
```
rayrutjes: Should this be writeHosts maybe?

rayrutjes: It's good as is

Nagriar: It's expected, this operation can be performed on a read-only node

PR #335, opened 2017-11-02
Replacing curl with Guzzle

garnold:
The PHP cURL extension [causes intermittent issues when running on Google App Engine](https://code.google.com/p/googleappengine/issues/detail?id=12451).  Guzzle provides an abstraction layer for HTTP requests, which allows you to change the underlying service without rewriting application code, and is [the recommended way to issue HTTP requests when using PHP on GAE](https://cloud.google.com/appengine/docs/php/issue-requests#issuing_an_http_request).

Guzzle will continue to use cURL if the extension is enabled, and fallback to stream handlers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,5 @@ vendor
 composer.lock
 .php_cs.cache
 tests/AlgoliaSearch/Tests/cache_dir/
+.idea/
+composer.phar
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Change Log
 
+### 2.0.0
+
+- feat: Replace curl with Guzzle (guzzlephp.org)
+
 ### 1.23.0
 
 - feat: add a `requestHeaders` parameter to every method to allow passing custom HTTP headers on a per request basis
--- a/composer.json
+++ b/composer.json
@@ -16,11 +16,16 @@
         {
             "name": "Ryan T. Catlin",
             "email": "ryan.catlin@gmail.com"
+        },
+        {
+            "name": "Geoffrey Arnold",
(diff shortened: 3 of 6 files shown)


julienbourdeau:
Hi @garnold,

Thank you for your contribution! 🙌 

Before we dive into the details, we need to discuss the PHP version. We are currenlty supporting PHP 5.3 but adding [Guzzle will require PHP 5.5](https://github.com/guzzle/guzzle#version-guidance).

rayrutjes:
Hey @garnold ,

Thanks a lot for your investigation and your PR.

Having an easy way to change the transport layer is definitely something we'd like to tackle at some point.
Even more when we hear about the issues you went through.

However, it has been a long journey to get the current cURL implementation work, be stable and be performant.

To change that part we would need to ensure 100% backward compatibility but also check the performance impact in different environments.

All in all, we definitely want to offer more flexibility, but there are also a lot of other things we need to work on that we would eventually want to release as part of a new major version.

As a result, for now I would suggest you keep a fork in sync with this repo.

I hope you understand my reasoning and that it makes sense.

garnold:
@julienbourdeau @rayrutjes Fine by me 👍 

Just to clarify, is it worth putting in the work to make this pull backwards compatible, or should I just hold off until you're ready to consider bumping PHP requirements through a major version release?

rayrutjes:
@garnold,

Moving this forward is not a high priority for us for now.
This might change in the near future though, but we have no ETA.

As a result, I recommend you keep your fork and work with it.

When we'll work on the second version, we'll make sure to integrate your work.

We maintain an internal list of all things that will need to be addressed, and making the transport layer more easier to fine tune is part of it.

Thanks for your work ❤️

PR #337, opened 2017-11-06
Fix/preserve bc

rayrutjes:


rayrutjes:
@alcaeus this should do the trick.

src/AlgoliaSearch/Client.php
```diff
@@ -235,8 +235,10 @@ public function setExtraHeader($key, $value)
      * @throws AlgoliaException
      * @throws \Exception
      */
-    public function multipleQueries($queries, $indexNameKey = 'indexName', $strategy = 'none', $requestHeaders = array())
```
alcaeus: Instead of removing the argument, I'd keep it in, but commented out. See https://github.com/symfony/symfony/blob/1e1b37753c1a6da4a56ede0e2fc9df19e1150ff1/src/Symfony/Component/DependencyInjection/Compiler/PassConfig.php#L121 for example.
Also, adding `@param array $requestHeaders` to the docblock will aid in IDEs understanding the variable signature and offering correct autocompletion.

julienbourdeau: Commenting the param might really help with discoverability. 
`@params` didn't help in my phpstorm. I don't know if there is some special configuration to set ¯\_(ツ)_/¯

alcaeus: I assumed it would, but a test revealed it didn't. Annotating a `@method` on the class doesn't help either, so at least PhpStorm gives precedence to the method signature.

rayrutjes:
@alcaeus you should now be able to use 1.23.1.
We left the commented parameters out for now in order to get something out quicker.

Please let us know if everything works for you now.

alcaeus:
@rayrutjes thanks for the release. I'll get back to you shortly.

alcaeus:
@rayrutjes works like a charm, thank you very much!

rayrutjes:
Awesome, thanks for jumping in @alcaeus .

PR #352, opened 2017-12-08
Introduce Scoped Copy Index

julienbourdeau:
Please note that for internal reason when scopedCopyIndex doesn't include `settings`, the engine will notre create the index. You have to create it manually before or after.

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -364,7 +364,33 @@ class Client
     {
         $requestHeaders = func_num_args() === 3 && is_array(func_get_arg(2)) ? func_get_arg(2) : array();
 
-        $request = array('operation' => 'copy', 'destination' => $dstIndexName);
+        return $this->scopedCopyIndex($srcIndexName, $dstIndexName, array(), $requestHeaders);
+    }
+
+    /**
+     * Copy an existing index and define what to copy along with records:
+     *  - settings
+     *  - synonyms
+     *  - query rules
+     *
+     * By default, everything is copied.
+     *
+     * @param string $srcIndexName the name of index to copy.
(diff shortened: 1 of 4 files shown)


src/AlgoliaSearch/Client.php
```diff
@@ -364,7 +364,33 @@ public function copyIndex($srcIndexName, $dstIndexName)
...
+     * @return mixed
+     */
+    public function scopedCopyIndex($srcIndexName, $dstIndexName, array $scope = array(), array $requestHeaders = array())
+    {
+        $request = array(
+            'operation' => 'copy',
+            'destination' => $dstIndexName,
+        );
+
+        if (! empty($scope)) {
```
maxiloc: Do we need this ? If the scope is [] then it might be ok

tests/AlgoliaSearch/Tests/MoveIndexTest.php
```diff
@@ -92,4 +96,76 @@ public function testCopyIndex()
         $this->assertEquals(1, $res['nbHits']);
         $this->assertEquals('Robin', $res['hits'][0]['firstname']);
     }
+
+    public function testScopedCopyIndex()
+    {
+        // Create source index with records, settings, synonyms and rules
```
maxiloc: It might be better to do several methods it would be easier to read, and understand what we are testing

src/AlgoliaSearch/Client.php
julienbourdeau: if you pass an empty scope with an empty array you get an error. As Clement puts it: "If you say it, you mean it".

maxiloc: 👍

tests/AlgoliaSearch/Tests/MoveIndexTest.php
julienbourdeau: Yes I should refactor them

PR #353, opened 2017-12-08
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertCount` instead of `count` function;
- `assertNotNull` instead of comparison with `null` keyword;
- `assertNotEquals` instead of comparisons `!=`;
- `assertGreaterThan` for mathematical comparions;
- `assertArrayHasKey` and `assertArrayNotHasKey` instead of `isset` function.


julienbourdeau:
Thank you!

PR #355, opened 2017-12-13
Fix tests for older php version

julienbourdeau:


tests/AlgoliaSearch/Tests/FileFailingHostsCacheTest.php
```diff
@@ -88,7 +85,7 @@ public function testThatTtlCanBeOverridden()
     /**
      * @param int $ttl
      *
-     * @return FailingHostsCache
+     * @return FileFailingHostsCache
```
rayrutjes: Not sure we need this change

julienbourdeau: Ah yes, you're right I didn't get it was the interface. I was mislead by PhpStorm ^^

julienbourdeau:
Tests are green 🎉

PR #357, opened 2017-12-18
Allow user to pass fully detailled key to `addApiKey`

julienbourdeau:
Currently if you do

```php
$client->addApiKey([
    'acl' => ['search'],
    'validity' => 24*60*60,
]);
```

The validity will not be set because the current method overrides it with default value of the second parameter.


julienbourdeau:
Spotted by @maxiloc :)

PR #376, opened 2018-03-01
Internal cleanup

julienbourdeau:
Some cleanup and variable renaming after the recent client review

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -247,7 +247,7 @@ class Index
      *
      * @throws \Exception
      */
-    public function getObjects($objectIDs, $attributesToRetrieve = null)
+    public function getObjects($objectIDs, $attributesToRetrieve = '')
     {
         $requestHeaders = func_num_args() === 3 && is_array(func_get_arg(2)) ? func_get_arg(2) : array();
 
@@ -459,7 +459,7 @@ class Index
         return $this->batch($requests, $requestHeaders);
     }
 
-    public function deleteBy(array $args)
+    public function deleteBy(array $filterParameters)
     {
         $requestHeaders = func_num_args() === 2 && is_array(func_get_arg(1)) ? func_get_arg(1) : array();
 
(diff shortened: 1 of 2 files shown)


src/AlgoliaSearch/Index.php
```diff
@@ -1476,14 +1474,6 @@ public function searchSynonyms($query, array $synonymType = array(), $page = nul
             $params['type'] = implode(',', $types);
         }
 
-        if ($page !== null) {
```
rayrutjes: What happens if called like this: `searchSynonyms('whatever', [], null, null)`?
Will the null values be striped out or ignored at some point afterwards?
If not, then this is a breaking change.

src/AlgoliaSearch/Index.php
```diff
@@ -1053,17 +1054,11 @@ public function setSettings($settings, $forwardToReplicas = false)
...
-            $url = $url.'?forwardToReplicas=true';
-        }
-
         return $this->client->request(
             $this->context,
             'PUT',
-            $url,
-            array(),
+            '/1/indexes/'.$this->urlIndexName.'/settings',
+            array('forwardToReplicas' => $forwardToReplicas),
```
rayrutjes: Are you sure this works?
Previously we had a string, here we give it a boolean instead.

src/AlgoliaSearch/Index.php
```diff
@@ -1744,9 +1736,11 @@ public function batchRules($rules, $forwardToReplicas = false, $clearExistingRul
         return $this->client->request(
             $this->context,
             'POST',
-            '/1/indexes/'.$this->urlIndexName.'/rules/batch?clearExistingRules='.($clearExistingRules ? 'true' : 'false')
-            .'&forwardToReplicas='.($forwardToReplicas ? 'true' : 'false'),
-            null,
+            '/1/indexes/'.$this->urlIndexName.'/rules/batch',
+            array(
+                'clearExistingRules' => $clearExistingRules,
```
rayrutjes: Same comment as for the forwardToReplicas.

src/AlgoliaSearch/Index.php
julienbourdeau: It worked but I only tested manually. I'll add proper tests 👍

src/AlgoliaSearch/Index.php
julienbourdeau: Good point! We replaced the detault value with the engine default value but I think if you call the method explicitly with `null`, I think it will fail. I'll try it.

julienbourdeau:
I removed the commit about forwardToReplicas, at this stage it is pointless to change it and could have side effects. Thanks Raymond for reviewing 👌

Issue #365, opened 2018-01-19
Algolia addObject method not working for floats.


luo-justin:
a simple add object method like this 

   $index->addObject(
              [
                 "avg_rating"=> 1.23456,
                'objectID' => '573'
            ]);

will result in an avg_rating of 0 on my algolia online records. However when i change avg_rating to 1 or 1.0, it is stored as 1. It seems that it cannot store any decimal point numbers. I am using the codeigniter framework. Does anyone else have that issue?

julienbourdeau:
Hi @chunkymankey,

Are you using another package, like a codigniter integration? I assume you are running a very recent version of the client (latest would be the best).

I have tested to be sure and he appear to work. Can you dig a bit deeper and see what is sent to Algolia [here](https://github.com/algolia/algoliasearch-client-php/blob/master/src/AlgoliaSearch/Client.php#L1239).

chunkymankey:
Hi, 

I echoed out the $body variable that you highlighted, and it outputs the following:

{"params":"query=Subject+Tutoring”}
{"avg_rating":1.2345600000000001017497197608463466167449951171875,"objectID":"573"}
{"avg_rating":1.2345600000000001017497197608463466167449951171875,"objectID":"573"}

This  is using the exact same method in my previous comment. It seems correct.

julienbourdeau:
I don't see how this would be converted to 0, for sure you should see `1.23456` in your dashboard.
About 1.0, it's expected to become 1.
Can you check in your logs (logs tab in your index in Algolia dashboard) what arrived to the API?

chunkymankey:
The response body updates correctly on the logs tab when i add an object:
{
  "avg_rating": 1.2345600000000001017497197608463466167449951171875,
  "objectID": "573"
}
However, the dashboard is still showing a 0. 
This is the JSON answer for object "573" on my dashboard:

{
  "hits": [
    {
      "avg_rating": 0,
      "objectID": "573",
      "_rankingInfo": {
        "nbTypos": 0,
        "firstMatchedWord": 0,
        "proximityDistance": 0,
        "userScore": 0,
        "geoDistance": 0,
        "geoPrecision": 1,
        "nbExactWords": 0,
        "words": 0,
        "filters": 1
      }
    }
  ],
  "nbHits": 1,
  "page": 0,
  "nbPages": 1,
  "hitsPerPage": 10,
  "processingTimeMS": 1,
  "facets": {},
[...]

PR #379, opened 2018-03-02
Check `serialize_precision` php setting

julienbourdeau:
See #365 for more details

This will add a tiny overhead to the instantiation of the client which I think is ok.

Ideally, I'd like to introduce a command like `composer doctor` that will check all the configuration and deps to avoid doing it in the code. This might be something we'll add to all API clients.

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -82,6 +82,9 @@ class Client
         if (!function_exists('json_decode')) {
             throw new \Exception('AlgoliaSearch requires the JSON PHP extension.');
         }
+        if (version_compare(PHP_VERSION, '7.1.0', '>') && ini_get('serialize_precision') !== '-1') {
+            @trigger_error('When using PHP 7.1+, you must set the "serialize_precision" ini settings to -1. See https://github.com/algolia/algoliasearch-client-php/issues/365', E_USER_WARNING);
+        }
 
         $this->caInfoPath = __DIR__.'/../../resources/ca-bundle.crt';
         foreach ($options as $option => $value) {


src/AlgoliaSearch/Client.php
```diff
@@ -82,6 +82,9 @@ public function __construct($applicationID, $apiKey, $hostsArray = null, $option
         if (!function_exists('json_decode')) {
             throw new \Exception('AlgoliaSearch requires the JSON PHP extension.');
         }
+        if (version_compare(PHP_VERSION, '7.1.0', '>') && ini_get('serialize_precision') !== '-1') {
+            throw new \Exception('When using PHP 7.1+, you must set the "serialize_precision" ini settings to -1. See https://github.com/algolia/algoliasearch-client-php/issues/365');
```
rayrutjes: Introducing a new Exception here might break existing implementations.
Also, I don't really think we should force the user to use a certain level of `serialize_precision`.
The application might already rely on another value for this configuration.

julienbourdeau: If they rely on PHP 7.1+ and another value, then their numbers are already broken. But yes, I'll change it to a `trigger_error` which won't stop the execution. Thanks for reviewing 🙌

rayrutjes: Numbers would only be broken if they use `float`s correct?

julienbourdeau: Yes `int`s work fine

rayrutjes:
I still don't agree even with triggering an error.
User might choose to keep `serialize_precision` to something else than `-1` and not use floats in any records. That is a valid use case and we should not force users to set any kind of globally used setting variable.
Even if they use floats, all it does is round to a different precision correct?

julienbourdeau:
Yes I see what you mean. I don't think anyone is using this parameter with PHP 7.1+. It does round differently, but I don't think anyone expect to transform 0.9 into 0.89999876 when json_encoding.

I agree it should be something we do in the constructor, it would be better if we had a script to launch to get debug information and recommendation (this `composer doctor` I had in mind).

rayrutjes:
I would definitely love to see this in the requirements check script 👍

PR #384, opened 2018-03-23
add curl requirement

OskarStark:
Fixes https://github.com/algolia/search-bundle/issues/217

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
     ],
     "require": {
         "php": ">=5.3",
-        "ext-mbstring": "*"
+        "ext-mbstring": "*",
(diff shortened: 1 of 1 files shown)


rayrutjes:
I'm not sure how I feel about this one.
It is technically good, I'm just afraid some people do not have cURL loaded in their PHP CLI and rely on that to install the dependencies so this might be annoying for them.

OskarStark:
> some people do not have cURL loaded in their PHP CLI 

they should use composer install `--ignore-platform-reqs` in this case, but to get it work the extension is needed...

rayrutjes:
Fair enough @OskarStark 👐

PR #386, opened 2018-03-28
"Remove" $objectIDKey

### LINKED ISSUE #369: Index Objects methods remove $objectIDKey args
* Index::addObjects
* Index::saveObjects
* Index::partialUpdateObjects

julienbourdeau:
Fix #369 

The parameter `$objectIDKey` only existed in the the PHP client. We decided to remove it to keep consistency with other client and ease documentation.

The signature of each method was kept for backward compatibility but should be used as if it was changed.

| Before                                                                                                     | After                                                                           |
|------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------|
| addObjects($objects, $objectIDKey = 'objectID', $requestOptions = [])                                      | addObjects($objects, $requestOptions = [])                                      |
| partialUpdateObjects($objects, $objectIDKey = 'objectID', $createIfNotExists = true, $requestOptions = []) | partialUpdateObjects($objects, $createIfNotExists = true, $requestOptions = []) |
| saveObject($object, $objectIDKey = 'objectID', $requestOptions = [])                                       | saveObject($object, $requestOptions = [])                                       |
| saveObjects($objects, $objectIDKey = 'objectID', $requestOptions = [])                                     | saveObjects($objects, $requestOptions = [])                                     |

## TODO

- [x] Remove `$objectIDKey` parameter without breaking backward compat'
- [x] Add tests for BC and new signature

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -172,9 +172,19 @@ class Index
      *
      * @return mixed
      */
-    public function addObjects($objects, $objectIDKey = 'objectID')
+    public function addObjects($objects, $objectIDKeyLegacy = 'objectID')
     {
-        $requestHeaders = func_num_args() === 3 && is_array(func_get_arg(2)) ? func_get_arg(2) : array();
+
+        $objectIDKey = 'objectID';
+        if (is_string($objectIDKeyLegacy)) {
+            $objectIDKey = $objectIDKeyLegacy;
+        }
+
+        $requestHeaders = array();
+        $nbArgs = func_num_args();
+        if ($nbArgs > 1) {
+            $requestHeaders = is_array(func_get_arg($nbArgs-1)) ? func_get_arg($nbArgs-1) : array();
+        }
 
         $requests = $this->buildBatch('addObject', $objects, true, $objectIDKey);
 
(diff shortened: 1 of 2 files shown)


src/AlgoliaSearch/Index.php
```diff
@@ -172,9 +172,19 @@ public function addObject($content, $objectID = null)
      *
      * @return mixed
      */
-    public function addObjects($objects, $objectIDKey = 'objectID')
+    public function addObjects($objects, $objectIDKeyLegacy = 'objectID')
     {
-        $requestHeaders = func_num_args() === 3 && is_array(func_get_arg(2)) ? func_get_arg(2) : array();
+
+        $objectIDKey = 'objectID';
+        if (is_string($objectIDKeyLegacy)) {
```
OskarStark: you should throw a deprecation here via `@trigger_error('message....', E_USER_DEPRECATED)` that this will be removed in the next major version.

src/AlgoliaSearch/Index.php
```diff
@@ -307,14 +317,27 @@ public function partialUpdateObject($partialObject, $createIfNotExists = true)
...
+     * @param string $objectIDKey       This parameter is deprecated and should not be used
      * @param bool   $createIfNotExists
      *
      * @return mixed
      */
-    public function partialUpdateObjects($objects, $objectIDKey = 'objectID', $createIfNotExists = true)
+    public function partialUpdateObjects($objects, $createIfNotExistsOrobjectIDKeyLegacy = 'objectID', $createIfNotExistsLegacy = true)
     {
-        $requestHeaders = func_num_args() === 4 && is_array(func_get_arg(3)) ? func_get_arg(3) : array();
+        if (is_bool($createIfNotExistsOrobjectIDKeyLegacy)) {
```
OskarStark: same here and below

src/AlgoliaSearch/Index.php
julienbourdeau: Thanks for reviewing!
We are currently discussing the futur of the client. There might be a next major where we remove everything that was deprecated but there might also be a new version fully redesign. In that case, it wouldn't make sense to throw deprecation notice.

PR #388, opened 2018-03-28
Build MCM tests for community PRs

julienbourdeau:
The issue with MCM was from somewhere else (see https://github.com/algolia/algoliasearch-client-php/commit/f53ae1bcf4b70927274d1989922fe182510c9dde), [the API Key dealer](https://blog.algolia.com/travis-encrypted-variables-external-contributions/) **should** provide a valid key to run these tests.


julienbourdeau:
MCM requires the actual admin key :(

PR #389, opened 2018-03-29
Add missing return statement in Client::isAlive() method

prestamodule:
Also, this method could maybe return a boolean value.
As we were not able to find the API doc for "isalive", we only added the return statement.

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -199,7 +199,7 @@ class Client
      */
     public function isAlive()
     {
-        $this->request(
+        return $this->request(
             $this->context,
             'GET',
             '/1/isalive',


julienbourdeau:
Thanks @prestamodule for your contribution.

I had a look and the client usually throw an exception when it cannot connect. So if you use it like `if ($client->isAlive())` it will break when not alive. I'd recommend to wrap the call with a try/catch instead, like other calls. More generally, I wouldn't worry about this method but wrap other method (like `addObjects` and such) with a try/catch.

Please let me know how that worked for you.

prestamodule:
`if ($client->isAlive())` cannot work in case it does not fire exception, because of the missing return statement. It will fail the condition because `$client->isAlive = void`

julienbourdeau:
Thanks for your contribution @prestamodule. Don't hesitate to send more PR if you think some things could be improved (with backward compat').

For this one, I'd rather not change it and rely on try/catch to check if Algolia is alive.

PR #391, opened 2018-04-04
Optimises method setConnectTimeout

nunomaduro:
Hey 👋

This PR addresses a small optimization on the method `setConnectTimeout` of the class `AlgoliaSearch \Client`. 

- Improves code readability.
- Improves **performance** avoiding methods such as `version_compare` and `curl_version` if the `$this->context->connectTimeout` is less than 1.

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -128,15 +128,18 @@ class Client
      */
     public function setConnectTimeout($connectTimeout, $timeout = 30, $searchTimeout = 5)
     {
-        $version = curl_version();
-        $isPhpOld = version_compare(phpversion(), '5.2.3', '<');
-        $isCurlOld = version_compare($version['version'], '7.16.2', '<');
+        if ($this->context->connectTimeout < 1) {
+            $version = curl_version();
+            $isPhpSupported = version_compare(phpversion(), '5.2.3', '>=');
+            $isCurlSupported = version_compare($version['version'], '7.16.2', '>=');
 
-        if (($isPhpOld || $isCurlOld) && $this->context->connectTimeout < 1) {
-            throw new AlgoliaException(
-                "The timeout can't be a float with a PHP version less than 5.2.3 or a curl version less than 7.16.2"
-            );
+            if (! $isPhpSupported || ! $isCurlSupported) {
+                throw new AlgoliaException(
+                    "The timeout can't be a float with a PHP version less than 5.2.3 or a curl version less than 7.16.2"
+                );
+            }
         }
+
         $this->context->connectTimeout = $connectTimeout;
         $this->context->readTimeout = $timeout;
         $this->context->searchTimeout = $searchTimeout;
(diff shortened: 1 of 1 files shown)


src/AlgoliaSearch/Client.php
```diff
@@ -128,15 +128,18 @@ public function __destruct()
...
-        $isPhpOld = version_compare(phpversion(), '5.2.3', '<');
-        $isCurlOld = version_compare($version['version'], '7.16.2', '<');
-
-        if (($isPhpOld || $isCurlOld) && $this->context->connectTimeout < 1) {
-            throw new AlgoliaException(
-                "The timeout can't be a float with a PHP version less than 5.2.3 or a curl version less than 7.16.2"
-            );
+        if ($this->context->connectTimeout < 1) {
+            $version = curl_version();
+            $isPhpSupported = version_compare(phpversion(), '5.2.3', '>=');
```
julienbourdeau: I'm thinking we should even remove this. We only support PHP 5.3+ and since we're using namespace, if you try to use the lib with 5.2, it will fail way before. WDYT?

src/AlgoliaSearch/Client.php
```diff
@@ -128,15 +128,18 @@ public function __destruct()
...
     {
-        $version = curl_version();
-        $isPhpOld = version_compare(phpversion(), '5.2.3', '<');
-        $isCurlOld = version_compare($version['version'], '7.16.2', '<');
-
-        if (($isPhpOld || $isCurlOld) && $this->context->connectTimeout < 1) {
-            throw new AlgoliaException(
-                "The timeout can't be a float with a PHP version less than 5.2.3 or a curl version less than 7.16.2"
-            );
+        if ($this->context->connectTimeout < 1) {
```
julienbourdeau: Isn't weird that we compare to `$this->context->connectTimeout` instead of `$timeout`. 🤔
I know it's not related to your PR and it was the case before, but it seems wrong to me.

nunomaduro: @julienbourdeau Ah, well spotted! 👍

On the current `setConnectTimeout` we are just validating the old `connectTimeout` that already exists on the `context`.

I will update the PR this weekend. We should also consider write a test for this particular case.

src/AlgoliaSearch/Client.php
nunomaduro: @julienbourdeau I think you are right! Namespaces are not supported prior to 5.3. The curl validation must remain tho. 👍

src/AlgoliaSearch/Client.php
```diff
@@ -128,15 +128,16 @@ public function __destruct()
      */
     public function setConnectTimeout($connectTimeout, $timeout = 30, $searchTimeout = 5)
     {
-        $version = curl_version();
-        $isPhpOld = version_compare(phpversion(), '5.2.3', '<');
-        $isCurlOld = version_compare($version['version'], '7.16.2', '<');
+        if ($connectTimeout < 1) {
```
julienbourdeau: The message says we cannot handle `float` with older verison of Curl, but the test only checks if it's less than 1. Should we check of `is_float` or is the message wrong? Or maybe 1.4 will be rounded to 1 interally and only the case with 0.x doesnt work because it's rounded to 0. 🤔

PR #392, opened 2018-04-06
UserAgent: Add language version

julienbourdeau:
I added a "lang" suffix to differenciate clearly between the package version and the language version.

### DIFF
--- a/src/AlgoliaSearch/Version.php
+++ b/src/AlgoliaSearch/Version.php
@@ -52,6 +52,10 @@ class Version
             self::$defaultUserAgentSegments =
                 'Algolia for PHP ('.self::VALUE.'); ' .
                 'PHP ('.$version.')';
+
+            if (defined('HHVM_VERSION')) {
+                self::$defaultUserAgentSegments .= '; HHVM ('.HHVM_VERSION.')';
+            }
         }
 
         $userAgent = self::$defaultUserAgentSegments . static::$suffixUserAgentSegments;
--- a/src/AlgoliaSearch/Version.php
+++ b/src/AlgoliaSearch/Version.php
@@ -33,7 +33,6 @@ class Version
 
     public static $custom_value = '';
 
-    private static $prefixUserAgentSegments = '';
     private static $defaultUserAgentSegments = '';
     private static $suffixUserAgentSegments = '';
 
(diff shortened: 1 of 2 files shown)


rayrutjes:
Did we think off the impact of this change in terms of monitoring?
This change diverges from our current specs. Did we update them?
Might be worth adding a test given this is called on every single Algolia operation.

julienbourdeau:
Yeah actually _lang_ doesn't make much sense, I'll remove it.
I'm waiting to see if we prefer the full version or the "simple version", we're not sure yet but I updated the tests.

julienbourdeau:
There is no rush really, but the best way to get the conversation moving was to open this PR and see how we can had this version. Most API clients don't have this suffix/prefix or even custom value and most languages don't have as many integration as php. So the PHP client is a good candidate for the first implementation.

About the spec, I don't see how this doesn't follow it. The spec doesn't state this order (unfortunately) and we consider the OS to be the platform.

Ideally, we'll need to update the UA in the integration when we update the client to remove the PHP version.
As far as I understand, the only thing that will happen is that we get a user agent with 2 entry about PHP, something like `PHP (7.1.3); ...; PHP (7.1.3-12ubuntuXXX)`.

rayrutjes:
> About the spec, I don't see how this doesn't follow it. The spec doesn't state this order (unfortunately) and we consider the OS to be the platform.

Seems like this sentence is contradictory. "I don't see how this doesn't follow it" vs "The spec doesn't state this order".
Anyway, I think the spec aren't very well enforced. However I think it is a good opportunity to freshen up the specs and apply them.

Regarding `PHP (7.1.3); ...; PHP (7.1.3-12ubuntuXXX)`, I just want to make sure that it doesn't for example double the amount of PHP usage we monitor. I don't think it would be double checking it would be interesting.

Feel free to dismiss my review once you guys decided on which way to go.
I just wanted to raise a few concern and make sure we thought about the impact.

aseure:
For the Top API clients usage that we list on customers dashboard, I confirm that this change won't have any impact.

Regarding the change itself, did we thought about using a `Language` (for instance) key instead of PHP? I'm saying that because it would need we would have a different key for every implementation, whereas I would have expected this to be implemented in all API clients with the same key name.

julienbourdeau:
I have checked in the dashboard and the client version must be passed first, I updated the code.

Please see https://gist.github.com/julienbourdeau/7c11df8647e85f5fc7eac9d2c017e52c

If the tests pass, the PR is ready to merge 🎉

PR #393, opened 2018-04-08
Use dedicated PHPUnit assertions

carusogabriel:
Following #353 :)


julienbourdeau:
Thanks :) I'm relaunching travis build as it looked like random failure.

PR #394, opened 2018-04-09
fix: Prevent saving a rule with an empty objectID

aseure:
Preventing users to shoot themselves in the foot, following the issue with discovered in the Go client (see https://github.com/algolia/algoliasearch-client-go/issues/397).

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -1782,6 +1782,10 @@ class Index
             $content['objectID'] = $objectID;
         }
 
+        if (! $content['objectID']) {
+            throw new AlgoliaException('Cannot save the rule because `objectID` must be set and non-empty.');
+        }
+
         return $this->client->request(
             $this->context,
             'PUT',
--- a/tests/AlgoliaSearch/Tests/RulesTest.php
+++ b/tests/AlgoliaSearch/Tests/RulesTest.php
@@ -41,6 +41,16 @@ class RulesTest extends AlgoliaSearchTestCase
     }
 
     /**
+     * @expectedException \AlgoliaSearch\AlgoliaException
+     */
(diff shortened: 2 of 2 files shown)


src/AlgoliaSearch/Index.php
```diff
@@ -1782,6 +1782,10 @@ public function saveRule($objectID, $content, $forwardToReplicas = false)
             $content['objectID'] = $objectID;
         }
 
+        if ($content['objectID'] !== '') {
```
julienbourdeau: I think you meant the opposite condition. Since we ensure the `objectID` key is set in `$content`, I'd simply check

```php
if (! $content['objectID']) {
    //
}
```

src/AlgoliaSearch/Index.php
```diff
@@ -1782,6 +1782,10 @@ public function saveRule($objectID, $content, $forwardToReplicas = false)
...
+            throw new AlgoliaException('Missing objectID.');
```
julienbourdeau: About the message, we could details explaining that the object ID must be set.

aseure:
@julienbourdeau Could you help me understanding why the tests do not pass with PHP `5.3`/`5.4`/`5.5` only?

src/AlgoliaSearch/Index.php
julienbourdeau: yes, anything falsy: number 0, empty string, null and such.

julienbourdeau:
@aseure PHP < 5.6 is using an old version of the phpunit framework. The method `expectException` didn't exist yet.

You need to do it with a tag in the comment like https://github.com/algolia/algoliasearch-client-php/pull/394/files#diff-7072b6041354c0bca3760b2df8f89580R54

aseure:
Ready again for review @julienbourdeau. Thanks for the explanation.

PR #397, opened 2018-04-13
fix(phpdoc): adapt phpdoc on search arguments

babaorum:
From my previous tests to search in an index.
When I only need to use the facets to search, and not the full query search, the `$query` parameter should be set to `null`, not an empty `string`.

### DIFF
--- a/src/AlgoliaSearch/Index.php
+++ b/src/AlgoliaSearch/Index.php
@@ -476,7 +476,7 @@ class Index
     /**
      * Search inside the index.
      *
-     * @param string $query the full text query
+     * @param string|null $query the full text query or null for disabling it
      * @param mixed $args (optional) if set, contains an associative array with query parameters:
      *                      - page: (integer) Pagination parameter used to select the page to retrieve.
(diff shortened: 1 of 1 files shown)


babaorum:
Yes, I just start using the feature yesterday, but when I passed a bunch of facets and numeric filters to filter my search with the query being set to `null`, I had the results I was expecting.
But when I passed, an empty string instead of `null`, no results were found.

Is it not an expected behavior from the api itself ? Or should an empty string be converted to null for the query parameter by the sdk ?

julienbourdeau:
You should get the exact same results with empty string and null. The issue must be from somewhere else but I cannot think of any reason.

I have checked and the only difference is that if you pass `null`, there is an empty query `&query=` appended to the url params.

Could you try to make two calls and look at what you find in your Algolia logs? See if you can spot anything weird.

julienbourdeau:
Did you have time to check it out?

babaorum:
Well,
I am really sorry.
I just check it out. I don't seem to be able to reproduce it. It must have been an error on my side at the time.

Thank you for your time

julienbourdeau:
You're welcome! Thanks for reporting the issue :)

PR #408, opened 2018-06-04
Introducing AB Testing

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no
| Need Doc update   | yes

## What was changed

Algolia introduced (in Beta) a new feature to create AB Test directly from your dashboard. This PR aims to make it easier to manage ABTests programatically.

I'd like to introduce a new _"first-class citizen object"_ , along with `Client` and `Index`, named `Analytics`. More method should be added to the Analytics object if we validate it.

The Analytics API doc can be found here: chttps://www.algolia.com/doc/rest-api/analytics/. 

## New methods

```php
$analytics->getABTests(); // It's not called `list`, because it's paginated
$analytics->addABTest($abTestDefinition); // It's not called `save` because you can't set the abtestID
$analytics->getABTest($abTestID);
$analytics->stopABTest($abTestID);
$analytics->deleteABTest($abTestID);
```

### Setup your index

```php
$client = new \AlgoliaSearch\Client(getenv('ALGOLIA_APP_ID'), getenv('ALGOLIA_API_KEY'));

$index = $client->initIndex('products');

$index->setSettings([
    'replicas' => [
        'products-alt'
    ],
    'customRanking' => [
        'desc(popularity)',
        'asc(price)',
    ]
], true);

$alt = $client->initIndex('products-alt');
$alt->setSettings([
    'customRanking' => [
        'desc(nb_comments_ratio)',
        'desc(popularity)',
    ]
]);

$index->addObjects($allMyProducts);
```

### Create AB Test scenario

```php
$analytics = $client->initAnalytics();

$response = $analytics->addABTest([
    "name" => $name,
    "variants" => [
        [
            "index" => "products",
            "trafficPercentage" => 70,
            "description" =>  "popular and cheaper",
        ], 
        [
            "index" => "products-alt",
            "trafficPercentage" => 30,
            "description" =>  "popular and cheaper",
        ],
    ],
    "endAt" =>  (new \DateTime('next month'))->format('Y-m-d\TH:i:s\Z'), // Notice the format
]);

// Wait until the task is complete if necessary
$analytics->waitTask($response['index'], $response['taskID']);
```

### Search (as you normally would)

Our engine takes care of forwarding the request to the index, according to the `trafficPercentage` defined.

```php
$index->search('joe');
```

## Note to other maintainers

* Remember to activate ABTest for your app used by Travis
* Most endpoints requires either `setSettings` or `Analytics` ACL so it should be testable with the API Key Dealer

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -716,6 +716,11 @@ class Client
         return new Index($this->context, $this, $indexName);
     }
 
+    public function initAnalytics()
+    {
+        return new Analytics($this);
+    }
+
     /**
      * List all existing API keys with their associated ACLs.
      *
@@ -1327,9 +1332,14 @@ class Client
             return;
         }
 
-        $answer = Json::decode($response, true);
         $context->releaseMHandle($curlHandle);
         curl_close($curlHandle);
+
+        if ($http_status == 204) {
+            return '';
+        }
+
+        $answer = Json::decode($response, true);
 
         if (intval($http_status / 100) == 4) {
             throw new AlgoliaException(isset($answer['message']) ? $answer['message'] : $http_status.' error', $http_status);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 ## Change Log
 
-### Unreleased
+### UNRELEASED
 
 <Contributors, please add your changes below this line>
+
+* Introduce AB Testing feature - PR [#408](https://github.com/algolia/algoliasearch-client-php/pull/#408)
+    List/Create/Stop/Delete AB Tests programmatically
+    Introduce new Analytics object, wrapper around the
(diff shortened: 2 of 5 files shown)


src/AlgoliaSearch/Client.php
```diff
@@ -1301,10 +1306,15 @@ public function doRequest(
             return;
         }
 
-        $answer = Json::decode($response, true);
```
julienbourdeau: Analytics often return no content, and passing it to the decode method will throw a "Malformed JSON" Exception.

julienbourdeau: the API was updated, we could either remove it or leave it (it's still some sort of bug fix)

julienbourdeau:
1__php__bash_

🤔

PR #414, opened 2018-07-03
Set json encode unescaped unicode as default

moebrowne:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | N/A
| Need Doc update   | no

## What was changed

The `Json::encode` method now has the `JSON_UNESCAPED_UNICODE` option enabled by default

## Why it was changed

The reason for this change is that when encoding multi byte strings the size of the JSON object is significantly larger and starts to cause errors with max record sizes.

For example:

`strlen('검색 엔진')` => 13
**pre-change**: `strlen(Json::encode('검색 엔진'))` => 27
**post-change**: `strlen(Json::encode('검색 엔진'))` => 15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 <Contributors, please add your changes below this line>
 
+* Sets the JSON encode method to not escape unicode characters by default
+
 
 ### 1.27.0
 
--- a/src/AlgoliaSearch/Json.php
+++ b/src/AlgoliaSearch/Json.php
@@ -7,7 +7,7 @@ namespace AlgoliaSearch;
  */
 class Json
 {
-    public static function encode($value, $options = 0)
+    public static function encode($value, $options = JSON_UNESCAPED_UNICODE)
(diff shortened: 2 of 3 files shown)


julienbourdeau:
I have to say my level of expertise in charsets is too limited, I need to have a deeper look to be sure there is no unexpected behavior. Feel free to share anything else you found about that.

julienbourdeau:
Hi @moebrowne 

I checked how we calculate the record size we json decode the payload to loop over each records. The engine will json encode each record individually, remove space between json attribute to make sure `{ "name" :      "value"}` becomes `{"name":"value"}` and then count how many bytes are used.

So the way you serialize it on your side won't affect the actual record size. In case you get errors from the API that you fixed with this parameters, please let me know and send an example, I'll look deeper.

Because this flag wasn't available in PHP 5.3, I'd rather not merge this PR.

Thank you for your contribution 🥇

PR #416, opened 2018-07-23
Describe actual behavior of scoped copy operation

greg0ire:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## What was changed

some comments

## Why it was changed

The comments imply that records will be copied if you specify a scope.
This is not the behavior I observe, and this is also not what is
documented on https://www.algolia.com/doc/rest-api/search/#copymove-index :

> Keep in mind that when you use the scopes parameter, you will no
longer be copying records (objects) but only the specified scopes.

It might have been true in the past, but it is currently not the case.

### DIFF
--- a/src/AlgoliaSearch/Client.php
+++ b/src/AlgoliaSearch/Client.php
@@ -404,7 +404,7 @@ class Client
     }
 
     /**
-     * Copy an existing index and define what to copy along with records:
+     * Copy an existing index and define what to copy instead of records:
      *  - settings
      *  - synonyms
      *  - query rules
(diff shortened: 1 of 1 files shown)


julienbourdeau:
Hi @greg0ire ! Thanks for your contribution. I'll relauch travis and then merge it.

If you have some time, I have worked on a new major version of this package and I would love to hear what you think of it! I wrote about it here: https://discourse.algolia.com/t/introducing-php-api-client-v2/5635/2

PR #420, opened 2018-08-20
Introduce BatchIterator

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | yes
| Need Doc update   | yes/no

## Describe your change

`Index::saveObjects` can now take a `BatchIterator` instead of just an array of object. Using this iterator, you can index millions of records in just one line. The API client create a new http call, for each batch given by the iterator.

See example in `tests/`.

**Note that the return type of `saveObjects` is now an array of response from the API (one entry per batch).**

If you want to reindex your entire database, you foten need to chunk it youself and call `Index::saveObjects` multiple time. With this abstract iterator, you can easily create an iterator to avoid loading everything in memory.

### DIFF
--- a/src/Index.php
+++ b/src/Index.php
@@ -4,6 +4,7 @@ namespace Algolia\AlgoliaSearch;
 
 use Algolia\AlgoliaSearch\Exceptions\TaskTooLongException;
 use Algolia\AlgoliaSearch\Interfaces\ClientConfigInterface;
+use Algolia\AlgoliaSearch\Iterators\AbstractBatchIterator;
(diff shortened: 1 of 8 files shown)


nunomaduro:
@julienbourdeau Can you show me an example for usage of this feature?

julienbourdeau:
`ArrayBatchIterator` is great, it makes things easier 👍 We'll need to rebase the branch before merge.

julienbourdeau:
After discussing it in details, we think there is something to do here, to generate batch but this solution isn't good enough yet. I'll try again later, probably after 2..0.0 stable is out.

PR #424, opened 2018-09-06
Refacto/merge config

julienbourdeau:
In the current version, we have `Config` for global (static) configuration and `ClientConfig` for the configuration of a client intance. As discussed this morning, we'd like to have only one (instance) config.

I have refactored the UserAgent, the timeouts and waitBeforeRetry config. I believe it makes much more sense, it was a good move.

Although, I'm stuck with the HttpClient builder method and the `VERSION` number. Where do you thing these should move?

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -36,8 +36,11 @@ class Client implements ClientInterface
 
     public static function create($appId = null, $apiKey = null)
     {
-        $config = new ClientConfig($appId, $apiKey);
+        return static::createWithConfig(ClientConfig::create($appId, $apiKey));
+    }
 
+    public static function createWithConfig(ClientConfig $config)
+    {
         $apiWrapper = new ApiWrapper(
             Config::getHttpClient(),
             $config
--- a/src/Client.php
+++ b/src/Client.php
@@ -18,11 +18,17 @@ class Client implements ClientInterface
      */
     protected $api;
 
+    /**
+     * @var ClientConfig
+     */
+    protected $config;
+
     protected static $client;
 
-    public function __construct(ApiWrapper $apiWrapper)
+    public function __construct(ApiWrapper $apiWrapper, ClientConfig $config)
     {
         $this->api = $apiWrapper;
+        $this->config = $config;
     }
 
     public static function get()
@@ -46,12 +52,12 @@ class Client implements ClientInterface
             $config
         );
 
-        return new static($apiWrapper);
+        return new static($apiWrapper, $config);
     }
 
     public function initIndex($indexName)
     {
-        return new Index($indexName, $this->api);
+        return new Index($indexName, $this->api, $this->config);
     }
 
     public function setExtraHeader($headerName, $headerValue)
--- a/src/Client.php
+++ b/src/Client.php
@@ -290,7 +290,8 @@ class Client implements ClientInterface
     public function waitForKeyAdded($key, $requestOptions = array())
     {
         $retry = 1;
-        $maxRetry = Config::$waitTaskRetry;
+        $maxRetry = $this->config->getWaitTaskMaxRetry();
+        $time = $this->config->getWaitTaskTimeBeforeRetry();
 
         do {
             try {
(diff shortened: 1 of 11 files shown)


src/Client.php
```diff
@@ -18,11 +18,17 @@ class Client implements ClientInterface
      */
     protected $api;
 
+    /**
+     * @var ClientConfig
```
nunomaduro: What do you think about using Fully Qualified Namespaces for phpdocs?

src/Index.php
```diff
@@ -19,12 +20,18 @@ class Index implements IndexInterface
...
      * @var ApiWrapper
      */
-    private $api;
+    protected $api;
 
-    public function __construct($indexName, ApiWrapper $apiWrapper)
+    /**
+     * @var ClientConfig
+     */
+    protected $config;
```
nunomaduro: There is a reason for this to be protected? Otherwise, I would default to private.

src/Support/ClientConfig.php
```diff
@@ -8,9 +8,27 @@ class ClientConfig
 {
     private $config;
 
-    public function __construct($appId = null, $apiKey = null)
+    private $defaultWaitTaskTimeBeforeRetry = 100000;
```
nunomaduro: I think we can improve the declaration of the variable above, for me, it's not clear the unit used. Is it milliseconds? Is it seconds? What do you think about suffix the unit in all of them? 🧐

src/Index.php
nunomaduro: There is a reason for this to be protected? Otherwise, I would default to private.

src/Client.php
nunomaduro: What do you think about using Fully Qualified Namespaces for phpdocs?

src/Support/ClientConfig.php
```diff
@@ -8,9 +8,27 @@ class ClientConfig
...
+    private $defaultWaitTaskMaxRetry = 30;
+
+    private $defaultReadTimeout = 5;
+    private $defaultWriteTimeout = 5;
+    private $defaultConnectTimeout = 2;
+
+    public function __construct(array $config = array())
+    {
+        $config += $this->getDefaultConfig();
```
nunomaduro: Union of $config and $this->getDefaultConfig() with preserve the `key => values` provided by $config in collision cases?

src/Support/ClientConfig.php
nunomaduro: Ok @julienbourdeau, just saw that those are options specified on the documentation. No need for units 👍

src/Support/UserAgent.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Support;
+
+final class UserAgent
```
nunomaduro: This class should be marked as `@internal`. Otherwise, clients may use this class.

src/Support/UserAgent.php
```diff
@@ -0,0 +1,53 @@
...
+{
+    private static $value;
+
+    private static $customSegments = array();
+
+    public static function get()
+    {
+        if (null === self::$value) {
```
nunomaduro: Nice cache strategy! 👍

src/Support/ClientConfig.php
julienbourdeau: Yes, it will only had the default values if the key doesn't exist in `$config`.

src/Support/UserAgent.php
julienbourdeau: Some users will need to use it, typically if you are building an integration, you should call `UserAgent::addCustomSegments(...)` to add details about your stack.

src/Index.php
julienbourdeau: I usually default everything to private but in this case, I think it's a good feature to be able to extend the Client and Index as they are the most important classes of the lib.
Typically, that's what we do with the CanaryClient and CanaryIndex. If you can't access `$config` and `$api`, you won't be able to do much.

I agree that this is a decision we should take carefully tho 👍  This could change before the final release.

src/Client.php
julienbourdeau: I haven't really bothered with phpdoc yet ebcause I think this can still change (use an interface for instance) but you're right, we should use FQN 👍

PR #425, opened 2018-09-07
Refacto/remove default values

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | YES  
| Need Doc update   | yes

## Describe your change

I have removed the default values for optional parameters. After discussing it yesterday, we decided thst we should keep the engine values if the user doesn't specify them.

⚠️ This means that `forwardToRepliccas` is default to FALSE (like in v1 and all other clients. The v2 originally defaulted it to TRUE, this is entirely rolled back.

There is one default value left, it's the `getVersion=2` for `Index::getSettings`. I believe this one still makes a lot of sense. The engine introduce the version for settings to keep BC but the version2 should be used as much as possible.

### DIFF
--- a/src/Index.php
+++ b/src/Index.php
@@ -60,13 +60,16 @@ class Index implements IndexInterface
 
     public function getSettings($requestOptions = array())
     {
+        if (is_array($requestOptions)) {
+            $requestOptions['getVersion'] = 2;
+        } elseif ($requestOptions instanceof RequestOptions) {
+            $requestOptions->addQueryParameter('getVersion', 2);
+        }
+
         return $this->api->read(
             'GET',
             api_path('/1/indexes/%s/settings', $this->indexName),
-            $requestOptions,
-            array(
-                'getVersion' => 2,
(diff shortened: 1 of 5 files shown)


src/Index.php
```diff
@@ -76,10 +76,7 @@ public function setSettings($settings, $requestOptions = array())
             'PUT',
             api_path('/1/indexes/%s/settings', $this->indexName),
             $settings,
-            $requestOptions,
-            array(
-                'forwardToReplicas' => true,
-            )
+            $requestOptions
```
nunomaduro: This is a change of behaviour. Can you add this to the changelog?

julienbourdeau:
`getVersion=2` is now forced for getSettings, just like it is in the v1 of the client 👍 

$default parameters have been removed from the APIWrapper methods but still exists in the RequestOptionsFactory::create* methods because people might need it when they create their own request parameters.

PR #430, opened 2018-09-10
Refactor retry strategy and host management

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | Not sure  
| Need Doc update   | no

## Describe your change

I have refactored the host management to implement the 5minutes timeout if an host is down and to reconnect to the main host as fast as possible.

Next, I'm going to add the cache on the file system to restore the state.

### Help

We need a better name for `ClusterHosts`.

### DIFF
--- a/src/RetryStrategy/ClusterHosts.php
+++ b/src/RetryStrategy/ClusterHosts.php
@@ -1,6 +1,6 @@
 <?php
 
-namespace Algolia\AlgoliaSearch\Internals;
+namespace Algolia\AlgoliaSearch\RetryStrategy;
 
 class ClusterHosts
 {
--- a/src/RetryStrategy/ClusterHosts.php
+++ b/src/RetryStrategy/ClusterHosts.php
@@ -8,7 +8,7 @@ class ClusterHosts
 
     private $write;
 
-    public function __construct(array $read, array $write)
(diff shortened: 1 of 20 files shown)


src/RetryStrategy/ClusterHosts.php
```diff
@@ -0,0 +1,101 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\RetryStrategy;
+
+class ClusterHosts
```
nunomaduro: Shouldn't this class marked as final and internal?

src/RetryStrategy/Host.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\RetryStrategy;
+
+class Host
```
nunomaduro: Shouldn't this class marked as final and internal?

julienbourdeau:
I'll upfate the PR for the conflicts 👍 

About `@internal` I wasn't entirely sure yet what to tag. I think it should be Host and HostCollection. About ClusterHost, I'll check :)

julienbourdeau:
We 'll add `@internals` and `final` when we review

PR #431, opened 2018-09-11
Feature/improve logging

nunomaduro:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | yes

**_Note: Another Pull Request will come up to modify logging params + exceptions params._**

## Describe your change

This PR rewrites(writes) all of package's existing logging system. With the new code, users are given the ability to configure their logging setup and gives much greater flexibility when it comes to third-party integration.

The client now integrates a logger, which allow you to get some information about the request lifecycle.

### Logger

The client now integrates a logger, which allow you to get some information about the request lifecycle.

1. You can enable/disable it via a static call.

```php
Logger::enable();
$index->addObjects(objects)
Logger::disable();
```

2. Or you can also define your own `PSR-3` Logger:

```php
// Make sure you define the logger implementation before creating an instance of `Client::class`.
LogManager::setLogger($myLogger);

$client = Client::create($appId, $apiKey);
$client->initIndex('index_name')->saveObjects($objects);
```


src/Log/Logger.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function log($level, $message, array $context = array())
+    {
+        if (self::$isEnabled) {
+            $logMessage = array(
+                'level' => $level,
+                'message' => $message,
+                'content' => $context,
```
julienbourdeau: `context`

julienbourdeau:
I understand why it's called LoggerManager, but in general I like to avoid the using `Manager`. Let's see if we can figure out a better name. http://wiki.c2.com/?DontNameClassesObjectManagerHandlerOrData

Also, it happens that you cant find one (see my `IndexManager` in the Symfony bundle for instance)

julienbourdeau:
I'm thinking that maybe we should even merge classes handling the httplayer and logger (and upcoming cache) inside the same class (just thinking outloud, maybe it's a bad idea)

src/Support/ClientConfig.php
```diff
@@ -150,4 +159,30 @@ public function setWaitTaskTimeBeforeRetry($time)
 
         return $this;
     }
+
+    /**
+     * Gets the default logger.
+     *
+     * @return \Psr\Log\LoggerInterface
+     */
+    public static function getDefaultLogger()
```
julienbourdeau: I was thinking that we would have a public api like:

```php
$clientConfig->setLogger($myPSRLogger);
$clientConfig->getLogger();
ClientConfig::setDefaultLogger($myPSRLogger);
```

This would avoid having a static call in ApiWrapper

src/Internals/ApiWrapper.php
```diff
@@ -29,14 +30,23 @@ class ApiWrapper
...
         ClientConfigInterface $config,
-        RequestOptionsFactory $RqstOptsFactory = null
+        RequestOptionsFactory $RqstOptsFactory = null,
+        LoggerInterface $logger = null
     ) {
         $this->http = $http;
         $this->config = $config;
-        $this->requestOptionsFactory = $RqstOptsFactory ? $RqstOptsFactory : new RequestOptionsFactory($config);
+        $this->requestOptionsFactory = $RqstOptsFactory ?: new RequestOptionsFactory($config);
+        $this->logger = $logger ?: ClientConfig::getDefaultLogger();
```
julienbourdeau: See my comment below. It would replace it to `$this->logger = $logger ?: $this->config->getLogger();`.
The point is that everywhere, you get the logger and the config makes sure you get one

src/Http/Guzzle6HttpClient.php
```diff
@@ -16,20 +16,19 @@
...
+     *
+     * @var \Algolia\AlgoliaSearch\Interfaces\ClientConfigInterface
+     */
+    private $config;
 
-        $this->client = $client;
-    }
+    private $client;
 
-    public static function createWithConfig(array $config)
```
julienbourdeau: 👍 👍 👍

src/Interfaces/ClientConfigInterface.php
```diff
@@ -23,4 +25,11 @@ public function getConnectTimeout();
     public function getWaitTaskMaxRetry();
 
     public function getWaitTaskTimeBeforeRetry();
+
+    /**
+     * Gets the logger instance of the object.
+     *
+     * @return \Psr\Log\LoggerInterface
+     */
+    public function getLogger();
```
julienbourdeau: Do you have to rewrite it here when you extend the LoggerAwareInterface? I dunno, I rarely extend interfaces ^^

src/Internals/ApiWrapper.php
```diff
@@ -36,7 +37,7 @@ public function __construct(
     ) {
         $this->http = $http;
         $this->config = $config;
-        $this->requestOptionsFactory = $RqstOptsFactory ? $RqstOptsFactory : new RequestOptionsFactory($config);
```
julienbourdeau: 👍

src/Support/ClientConfig.php
```diff
@@ -150,4 +166,46 @@ public function setWaitTaskTimeBeforeRetry($time)
...
+    {
+        $this->logger = $logger;
+    }
+
+    /**
+     * Gets the default logger.
+     *
+     * @return \Psr\Log\LoggerInterface
+     */
+    public static function getDefaultLogger()
```
julienbourdeau: I'm not sure what the usecase would be for this method

src/Support/HttpLayer.php
```diff
@@ -2,25 +2,27 @@
...
 namespace Algolia\AlgoliaSearch\Support;
 
+use Algolia\AlgoliaSearch\Interfaces\ClientConfigInterface;
+
 final class HttpLayer
 {
     private static $httpClientConstructor;
 
-    public static function get()
+    public static function get(ClientConfigInterface $config)
```
julienbourdeau: 👍

src/Interfaces/ClientConfigInterface.php
nunomaduro: LoggerAwareInterface only haves setLogger, i don't have a getLogger.

PR #434, opened 2018-09-13
Introduce copySynonyms, copySettings, copyRules

julienbourdeau:
Discussed many times, I realized they weren't implemented. 😓

### DIFF
--- a/src/Client.php
+++ b/src/Client.php
@@ -171,7 +171,7 @@ class Client implements ClientInterface
             $requestOptions->addBodyParameter('scope', array('settings'));
         }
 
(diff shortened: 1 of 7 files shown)


src/Client.php
```diff
@@ -164,6 +163,39 @@ public function deleteIndex($indexName, $requestOptions = array())
         );
     }
 
+    public function copySettings($srcIndexName, $dstIndexName, $requestOptions = array())
```
nunomaduro: Shouldn't we add tests for those new features?

julienbourdeau:
@nunomaduro I have a added a new test suite to check what the request looks like before we send it to Algolia. Once it's merged, I'll rewrite `forwardToReplicas` tests to use it.

PR #437, opened 2018-09-14
[RFC] Introduce setSettingsWithReplicas() method

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | no

# Update

Instead of adding so many method, I introduce a config option for default to `forwardToReplica = true` based on @nunomaduro's feedback.

This will be used this way:

```php
$client = Client::createWithConfig(['defaultForwardToReplicas' => true]);

$client->initIndex('name')->setSettings($settings); // Will be forwarded to replicas
$client->initIndex('name')->setSettings($settings, ['fowardToReplicas' => false); // Will *NOT* be forwarded 

```

## Describe your change

~I added a method named `setSettingsWithReplicas` with basically call `setSettings` with the `forwardToReplicas` set to true (default is false).~

```php
$index->setSettingsWithReplicas($settings);
// Instead of 
$index->setSettings($settings, ['forwardToReplicas' => true]);
```

### Why

We know that most of the time, developer need to forward the settings, synonyms and rules to replica indexes.

When we started this v2, we decided to turn forwardToReplicas to `true`by default. We recently revised that and decided that we should stick to the REST API defaults. Which I think was the best decision.

The goal of this method is to make things explicit. Users will understand better that calling `setSettings` will not set the settings to replica indexes.

I believe the maintenance cost is very limited, the method only proxy another method, once it's written, everything should be fine.

### Tests

To test all the proxy methods, I'll add something to test that a proxy method call the correct mehtod with the correct arguments. See #434 

## Need feedback

I'm not sure this PR should be merge, I'd like to get your opinion on it. 🙏 

If we decide to move forward with this, we should also add:

```
saveSynonymWithReplicas()
saveSynonymsWithReplicas()
freshSynonymWithReplicas()
freshSynonymsWithReplicas()
clearSynonymWithReplicas()
clearSynonymsWithReplicas()
deleteSynonymWithReplicas()
deleteSynonymsWithReplicas()

saveRuleWithReplicas()
saveRulesWithReplicas()
freshRuleWithReplicas()
freshRulesWithReplicas()
clearRuleWithReplicas()
clearRulesWithReplicas()
deleteRuleWithReplicas()
deleteRulesWithReplicas()
```

### DIFF
--- a/src/Index.php
+++ b/src/Index.php
@@ -81,11 +81,17 @@ class Index implements IndexInterface
 
     public function setSettings($settings, $requestOptions = array())
     {
+        $default = array();
+        if (is_bool($fwd = $this->config->getDefaultForwardToReplicas())) {
+            $default['forwardToReplicas'] = $fwd;
+        }
+
         return $this->api->write(
             'PUT',
             api_path('/1/indexes/%s/settings', $this->indexName),
             $settings,
-            $requestOptions
+            $requestOptions,
+            $default
         );
     }
 
@@ -264,11 +270,17 @@ class Index implements IndexInterface
     {
         Helpers::ensureObjectID($synonyms, 'All synonyms must have an unique objectID to be valid');
 
+        $default = array();
+        if (is_bool($fwd = $this->config->getDefaultForwardToReplicas())) {
+            $default['forwardToReplicas'] = $fwd;
+        }
+
         return $this->api->write(
             'POST',
             api_path('/1/indexes/%s/synonyms/batch', $this->indexName),
             $synonyms,
-            $requestOptions
+            $requestOptions,
+            $default
         );
     }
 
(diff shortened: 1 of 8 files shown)


nunomaduro:
Nice proposal @julienbourdeau. 🎉

This change would definitely help users understand that previous methods, such as `clearRules`, **do not** `forwardToReplicas` to replicas.  👍

Concerning the developer experience, would prefer something such as: 
```php
$client->withReplicas();

$client->setSettings($settings);
$client->setRules($rules);
```

Or:
```php
ClientConfig::setForwardToReplicas(true);

$client->setSettings($settings);
$client->setRules($rules);
```

You know clients better me, by default what clients want? Always `forwardToReplicas` as true?

rayrutjes:
Hey, great RFC indeed.

I agree with @nunomaduro on this one. Not a huge fan of the naming `...WithReplicas`. 
What we really mean is "Set the settings on the current index and also on the replicas".
Otherwise said, we want to "propagate" the changes to replicas. That is why I think the `With` could be confusing.

I would probably go with a design similar to what @nunomaduro suggests, but I would try to find another naming. 

Quick brainstorming:
- (Set)PropagateChangesToReplicas
- (Set)PropagateToReplicas
- (Set)SyncWithReplicas
- (Set)SyncReplicas
- (Set)IncludeReplicas
- (Set)ExcludeReplicas

julienbourdeau:
I like having a global config for that, it's easy to document and you always know the state your in. If you had it via a setter, it might be harder to know if it's on or off, and it will be different for each client instance. This could also be seen as a feture, but I see it as a source of bug 😄

PR #440, opened 2018-09-17
[RFC] Introduce Relevancy helpers

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Need Doc update   | if merged

I'm thinking our client could go further and help with you relevancy.

If you want to sort by rating, you shouldn't sort by "average rating" or "number of ratings", you should use a ratio that takes both into account. Otherwise a product 1 one rating 5/5 will be shown before a product with 1000 rating but 4.9/5 on average.

You can read more about it here: http://www.evanmiller.org/how-not-to-sort-by-average-rating.html

Because Algolia is a tie breaking algorithm, we should also round this number to create ties! (Not implemented in this PR)

Please share your thoughts. Note that this probably won't make it in the 2.0.0 release.

### DIFF
--- a/src/Relevancy/Ratings.php
+++ b/src/Relevancy/Ratings.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Relevancy;
+
+class Ratings
+{
+    public static function score($positive, $negative) {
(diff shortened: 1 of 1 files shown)


rayrutjes:
I'm wondering if this is a good fit for the API client itself.
Seems like a good fit for a nice and clean dedicated (community) package.

shirshak55:
yes ratio based should be done :) Otherwise search is not good 👍

PR #443, opened 2018-09-24
Introduce ResponseObjects

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | yes if you use `is_array` on indexing response
| Need Doc update   | yes

## Describe your change

Every indexing operations returns an object with a `wait()` method. The response from the API is **still accessible as an array**.

```php
$response = $index->saveSynonym($synonym);
$response.wait();
echo $response['objectID'];

$response = $cliebt->addApiKey();
$response.wait();
echo $response['value'];
```

It makes waiting pretty cool. It's now consistant, whether you're waiting for a new API key or a standard indexing operations.

## To be tested

In order to wait, I inject the client or the index inside the response. I'm afraid of the consequenses on memory consumption. 
I unset the client or index at the end of a successful `wait()` call, as a small optimisation.

I'm definitely open to feedback on this one 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,15 @@
     
 * Remove Canary Client and Index
 
+* Introduce ResponseObjects
+    Every indexing operations returns an object with a `wait()` method.
+    The response from the API is still accessible as an array.
+    ```php
+    $response = $index->saveSynonym($synonym);
+    $response->wait();
+    echo $response['objectID'];
+    ```
+
 
 ### UNRELEASED
 
--- a/src/Client.php
+++ b/src/Client.php
@@ -2,8 +2,6 @@
 
 namespace Algolia\AlgoliaSearch;
 
-use Algolia\AlgoliaSearch\Exceptions\NotFoundException;
-use Algolia\AlgoliaSearch\Exceptions\TaskTooLongException;
 use Algolia\AlgoliaSearch\Http\HttpClientFactory;
 use Algolia\AlgoliaSearch\Interfaces\ConfigInterface;
 use Algolia\AlgoliaSearch\Interfaces\ClientInterface;
--- a/src/Client.php
+++ b/src/Client.php
@@ -7,6 +7,9 @@ use Algolia\AlgoliaSearch\Exceptions\TaskTooLongException;
 use Algolia\AlgoliaSearch\Http\HttpClientFactory;
 use Algolia\AlgoliaSearch\Interfaces\ConfigInterface;
 use Algolia\AlgoliaSearch\Interfaces\ClientInterface;
+use Algolia\AlgoliaSearch\Response\DeleteApiKeyResponse;
+use Algolia\AlgoliaSearch\Response\MultipleIndexingResponse;
+use Algolia\AlgoliaSearch\Response\AddApiKeyResponse;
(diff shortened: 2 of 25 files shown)


src/Response/IndexingResponse.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Response;
+
+use Algolia\AlgoliaSearch\Exceptions\CannotWaitException;
+use Algolia\AlgoliaSearch\Interfaces\IndexInterface;
+
+class IndexingResponse extends AbstractResponse
```
nunomaduro: All the responses are indexing operations? 🧐

tests/Unit/ResponseObjectTest.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    public function testResponseObjectIsArrayAccessible()
+    {
+        $res = new Response(array(
+            'value' => sha1('unit test'),
+            'taskId' => 345646787,
+            'something' => array('one', 'two'),
+        ));
+
+        dump($res['valuwefee']);
```
nunomaduro: Is this dump normal?

CHANGELOG.md
```diff
@@ -27,6 +27,15 @@
     
 * Remove Canary Client and Index
 
+* Introduce ResponseObjects
+    Every indexing operations returns an object with a `wait()` method.
+    The response from the API is still accessible as an array.
+    ```php
+    $response = $index->saveSynonym($synonym);
+    $response.wait();
```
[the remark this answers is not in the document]
julienbourdeau: oops 😅

src/Response/DeleteKeyResponse.php
```diff
@@ -0,0 +1,56 @@
...
+        $retry = 1;
+        $maxRetry = $this->config->getWaitTaskMaxRetry();
+        $time = $this->config->getWaitTaskTimeBeforeRetry();
+
+        do {
+            try {
+                $this->client->getApiKey($key, $requestOptions);
+
+            } catch (NotFoundException $e) {
+                unset($this->client, $this->config);
```
maxiloc: Are you sure this works. because I think $this->client is a reference to client instantiated by the user. So we might make the client unusable.

nunomaduro: @julienbourdeau Yeah, you may want to run a quick test on this. If doesn't work, there is always the `clone` option.

src/Client.php
```diff
@@ -208,7 +213,9 @@ public function getApiKey($key, $requestOptions = array())
 
     public function addApiKey($keyParams, $requestOptions = array())
     {
-        return $this->api->write('POST', api_path('/1/keys'), $keyParams, $requestOptions);
+        $response = $this->api->write('POST', api_path('/1/keys'), $keyParams, $requestOptions);
+
+        return new NewKeyResponse($response, $this, $this->config);
```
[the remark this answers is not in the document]
julienbourdeau: yes it's more consistent!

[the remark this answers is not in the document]
julienbourdeau: I would have regretted this naming for years!

PR #447, opened 2018-09-26
Refacto/config

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

I have extracted the Logger to `Algolia` class, so it works the same as the cache. I'm not sure holding the logger should be the responsability of the config class. Via Algolia class, you have to use the same logger for all your instance of Client/Index/Analytics/places.

I also introduce SimpleConfig, which has less option than ClientConfig for other apis.

We shouldn't have to pass a `ClientConfig` to `Analytics`, `Places`... This way, we can inject `SimpleConfig` which has way less config.


src/Config/ClientConfig.php
```diff
@@ -0,0 +1,81 @@
...
+
+    public function getDefaultForwardToReplicas()
+    {
+        return $this->config['defaultForwardToReplicas'];
+    }
+
+    public function setDefaultForwardToReplicas($default)
+    {
+        if (!is_bool($default)) {
+            throw new \InvalidArgumentException('Default configuration for ForwardToReplicas must be a boolean');
```
nunomaduro: I do love strict asserts. But we are not applying this treatment to other setters, such as `setWaitTaskTimeBeforeRetry`, etc.

Any plans to add it later? 🙂

src/RetryStrategy/ApiWrapper.php
```diff
@@ -181,6 +182,6 @@ public function setExtraHeader($headerName, $headerValue)
      */
     private function log($level, $message, array $context = array())
     {
-        $this->config->getLogger()->log($level, 'Algolia API client: '.$message, $context);
+        Algolia::getLogger()->log($level, 'Algolia API client: '.$message, $context);
```
nunomaduro: 🎉

src/Config/ClientConfig.php
julienbourdeau: If possible yes, do you have any other example in mind? It's not always easy because some param have fun values like distinct. But if we can, we should add this type of check.

src/Algolia.php
```diff
@@ -40,12 +49,36 @@ public static function getCache()
...
     }
+
+    /**
+     * Gets the logger instance
+     *
+     * @return \Psr\Log\LoggerInterface
+     */
+    public static function getLogger()
+    {
+        if (!self::$logger) {
```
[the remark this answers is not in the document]
julienbourdeau: I'll change cache at the same time

src/Config/ClientConfig.php
julienbourdeau: Sorry I didn't realize `setWaitTaskTimeBeforeRetry` was an example ^^ yes it should be added

PR #456, opened 2018-10-01
Introduce PlaceConfig and AnalyticsConfig instead of Simple config

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | 🤸‍♂️  
| Need Doc update   | no

## Describe your change

I recently added SimpleConfig along with ClientConfig. After giving it more thoughts, I think it was a mistake, I'd rather have a different config for each API. We'll be adding the monitoring API soon, so we can have "MonitoringConfig". I think this will be better in the long run.


julienbourdeau:
Daaamn seeing this build being green again feels amazing ❤️

PR #457, opened 2018-10-01
HttpClientInterface only requires sendRequest method!

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | YES - httpClient Interface changed     
| Need Doc update   | yes/no

## Describe your change

When I wrote the `HttpClientInterface`, I added 3 methods, but createUri and createRequest never belonged there, I just got confused because they were provided by guzzle and that I had to maintain a different implementation for PHP 5.3

In the end createUri and createRequest have become as private methods of `ApiWrapper` 🎉

### DIFF
--- a/src/Http/HttpClientInterface.php
+++ b/src/Http/HttpClientInterface.php
@@ -6,15 +6,5 @@ use Psr\Http\Message\RequestInterface;
 
 interface HttpClientInterface
 {
-    public function createUri($uri);
-
-    public function createRequest(
-        $method,
-        $uri,
-        array $headers = array(),
-        $body = null,
-        $protocolVersion = '1.1'
-    );
-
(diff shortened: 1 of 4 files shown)


src/Http/HttpClientInterface.php
```diff
@@ -6,15 +6,5 @@
...
-
-    public function createRequest(
-        $method,
-        $uri,
-        array $headers = array(),
-        $body = null,
-        $protocolVersion = '1.1'
-    );
-
     public function sendRequest(RequestInterface $request, $timeout, $connectTimeout);
```
nunomaduro: Be sure (or in another pr) you document via `phpdocs` the exceptions that may be throw on this interface.

PR #459, opened 2018-10-03
Move HttpClient management under Algolia class

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | yes     
| Need Doc update   | yes

## Describe your change

`Algolia` is used a some sort of container. It manages what's global to the library: cache, log and Http layer. In a full featured framework, this would be the job the container.

It made more sense to move the HttpClientFactory under the Algolia class


src/Algolia.php
```diff
@@ -81,4 +86,37 @@ public static function setLogger(LoggerInterface $logger)
...
+                self::setHttpClient(function () {
+                    return new \Algolia\AlgoliaSearch\Http\Php53HttpClient();
+                });
+            }
+        }
+
+        return forward_static_call(self::$httpClientConstructor);
+    }
+
+    public static function setHttpClient($httpClientConstructor)
```
nunomaduro: I would prefer set a `$httpClient` instance instead of a `callable`. There is a reason for this?

julienbourdeau: Hmmm, very good question.

Do we want to have a fresh HttpClient instance for each Client/Place/Analytics class, or do we want to use the same shared instance 🤔 

I actually don't know, we need to figure it out.

nunomaduro: @julienbourdeau You can always `clone` it every time you do a `Algolia::getHttpClient()`.

```php
public function getHttpClient()
{
    return clone $this->client;
}
```

PR #460, opened 2018-10-03
Install Guzzle if platform is compatible

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

After composer update, this script will check the current setup and install guzzle if PHP > 5 and no `LEGACY_HTTP` env var was found.

We cannot require guzzle because we need to support 5.3. I found this solution better and easier to maintain than having a separate package with HttpClient adapters and different dependencies.
see: https://stackoverflow.com/questions/51333587/php-composer-require-dependency-if-condition

### DIFF
--- a/bin/algolia-deps
+++ b/bin/algolia-deps
@@ -1,0 +1,44 @@
+#!/usr/bin/env php
+<?php
+
+/**
+ * This script will automatically require guzzle library if your
+ * setup is compatible.
+ * If you don't want to use it, you can remove it with
+ *  `composer remove guzzlehttp/guzzle`
+ *
+ * Setting a LEGACY_HTTP=true in your environment variable will
+ * prevent the installation of guzzle.
+ *
+ * Note that it's *HIGHLY* recommended to use it.
+ */
+
+if (getenv('LEGACY_HTTP') || PHP_VERSION_ID < 50500) {
+    exit(0);
+}
+
+if ('composer.json' !== $COMPOSER_JSON = getenv('COMPOSER') ?: 'composer.json') {
+    putenv('COMPOSER=composer.json');
+    $_SERVER['COMPOSER'] = $_ENV['COMPOSER'] = 'composer.json';
+}
+
+$root = __DIR__;
+while (!file_exists($root.'/'.$COMPOSER_JSON) || file_exists($root.'vendor/bin/algolia-doctor')) {
+    if ($root === dirname($root)) {
+        break;
+    }
+    $root = dirname($root);
+}
+dump($root);
+
+$oldPwd = getcwd();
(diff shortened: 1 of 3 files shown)


bin/algolia-deps
```diff
@@ -0,0 +1,44 @@
...
+$PHP = 'php';
+if ('phpdbg' === PHP_SAPI) {
+    $PHP .= ' -qrr';
+}
+
+$COMPOSER = file_exists($COMPOSER = $oldPwd.'/composer.phar') || ($COMPOSER = rtrim('\\' === DIRECTORY_SEPARATOR ? preg_replace('/[\r\n].*/', '', `where.exe composer.phar`) : `which composer.phar 2> /dev/null`))
+    ? $PHP.' '.escapeshellarg($COMPOSER)
+    : 'composer';
+
+passthru("$COMPOSER require guzzlehttp/guzzle");
```
nunomaduro: Do we support all guzzle versions? Otherwise you should specify it. Example: `composer require guzzlehttp/guzzle "^6.0"`

bin/algolia-deps
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env php
```
nunomaduro: May sounds silly idea: There is a way of testing this script? 😅

bin/algolia-deps
```diff
@@ -0,0 +1,44 @@
...
+<?php
```
nunomaduro: There is a way of performing a question on this script? I would prefer a question where the user can opt to install guzzle.

bin/algolia-deps
nunomaduro: Small feedback: Make sure the exit code and the output is respected in this method `passthru`.

bin/algolia-deps
```diff
@@ -0,0 +1,44 @@
...
+}
+
+$root = __DIR__;
+while (!file_exists($root.'/'.$COMPOSER_JSON) || file_exists($root.'vendor/bin/algolia-doctor')) {
+    if ($root === dirname($root)) {
+        break;
+    }
+    $root = dirname($root);
+}
+dump($root);
```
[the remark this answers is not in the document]
julienbourdeau: 🤣

julienbourdeau:
Thank you @nunomaduro for your review!

In the end, I'm closing this PR, we won't try to install guzzle automatically. It seems very hacky and error-prone. Plus, Composer will only excecute scripts of the root package, not dependencies'.

Maybe the best solution is the initial one, having a second package with http adapters. ¯\_(ツ)_/¯

PR #465, opened 2018-10-15
Added: new Index::rename method

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | yes/no

## Describe your change

Introduce rename method, which allows you to rename an index and avoid instanciating 2 objects:

```php
// Before
$index = $client->initIndex('name');
$index->saveObjects($objects);
$client->moveIndex('name', 'new_name');
$newIndex = $client->initIndex('new_name');
$newIndex->getSettings();

// After
$index = $client->initIndex('name');
$index->saveObjects($objects);
$index->rename('new_name');
$index->getSettings(); // will taget 'new_index'
```


tests/Unit/ResponseObjectTest.php
```diff
@@ -28,6 +28,8 @@ public function testIndexingResponse()
         $i = static::$client->initIndex('cool');
```
nunomaduro: Shouldn't we consider add a new test for the `rename` method? 🧐

src/Index.php
```diff
@@ -66,6 +68,23 @@ public function clear($requestOptions = array())
...
+            'POST',
+            api_path('/1/indexes/%s/operation', $this->indexName),
+            array(
+                'operation' => 'move',
+                'destination' => $newIndexName,
+            ),
+            $requestOptions
+        );
+
+        $this->setIndexName($newIndexName);
```
nunomaduro: Nice! Keeping the same object is a real cool experience! 👏

tests/Unit/ResponseObjectTest.php
nunomaduro: Another feedback: Would be good add the phpdoc: `@deprecated` to the method moveIndex.

maxiloc: No strong opinion on the deprecation. Your call.

Otherwise good to go

julienbourdeau: I quite like it, I'd rather keep it.
I'm thinking about when you write a clean up script, or you're going live and you rename things.

PR #467, opened 2018-10-17
Prepare release

julienbourdeau:
Edits and small fixes before release.


docs/UPGRADE-from-v1-to-v2.md
```diff
@@ -1,32 +1,171 @@
...
-### ApiKeys can only be managed by the client, not the index
+It is recommend to go over your Algolia implementation and check if the method signature has changed
 
-This was already deprecated and has been removed in v2
+✅ Unchanged
+🤞 Changed but similar
+🛑 Requires attention
 
-### Method signature change
+Note that the v1 had a magic `$requestHeaders` which as became part of `$requestOptions`. The argument was magic
```
rayrutjes: s/as/has

rayrutjes: Actually it should either be `has become` or `became` without as.

docs/UPGRADE-from-v1-to-v2.md
```diff
@@ -1,32 +1,171 @@
...
+so it's not used a lot, it's ignored for clarity purpose.
```
rayrutjes: I don't think we should justify too much here.
We can remove this sentence IMHO.

PR #470, opened 2018-10-24
ReplaceAll, Reindex, MigrateIndex

julienbourdeau:
- [x] Rename `fresh*` to `replaceAll*`
- [x] Add `replaceAllObjects`
    - copy Settings, Synonyms and Rules from prod index
    - has optional second param to waitTask (false by default)
- [x] `addObjects` can take any iterator and create batches of 1,000 objects.
- [x] batch size can be modified via the ClientConfig
- [x] Add `reindex` method (see example below)
    - Introduce `IndexContentInterface` 
    - has optional second param to waitTask (false by default)
    - REPLICAS? 🤷‍♂️
- [ ] Add `Client::migrateIndex`

### Reindex method

IndexContent classes are a way to represent what to send in an index. It must implement the `IndexContentInterface`. You must provide objects, but settings/synonyms/rules are optional. 

If `getSettings`, `getSynonyms` or `getRules` return `false`, the resources will be copied from the production index.

```php
use Algolia\AlgoliaSearch\Interfaces\IndexContentInterface;

class ContactIndexContent implements IndexContentInterface
{

    public function getObjects()
    {
        return \Contact::getAll();
    }

    public function getSettings()
    {
        return array(
            'hitsPerPage' => 1000,
            'searchableAttributes' => array('firstname', 'lastname', 'surname', 'address'),
            'customRanking' => array('reputation'),
        );
    }

    public function getSynonyms()
    {
        return json_decode(
            file_get_contents(__DIR__.'/contact-synonyms.json'),
            true
        );
    }

    public function getRules()
    {
        // If you manage rules via the dashboard, return `false`
        // so it will be copied from the production index
        return false;
    }
}
```


src/Interfaces/SearchClientInterface.php
```diff
@@ -2,7 +2,7 @@
 
 namespace Algolia\AlgoliaSearch\Interfaces;
 
-interface ClientInterface
+interface SearchClientInterface
```
rayrutjes: How do we plan to introduce new endpoints in a backward compatible way?

julienbourdeau: Thank you @rayrutjes 🙏 
I don't know why I added these interfaces, I don't think they bring any value! I'll remove them and have look at the other interfaces in the meantime.

src/SearchIndex.php
```diff
@@ -170,6 +171,33 @@ public function saveObject($object, $requestOptions = array())
...
+            unset($requestOptions['objectIDKey']);
+        }
+
+        try {
+            return $this->splitIntoBatches('updateObject', $objects, $requestOptions);
+        } catch (MissingObjectId $e) {
+            $message = "\nAll objects must have an unique objectID (like a primary key) to be valid.\n\n";
+            $message .= "If your batch has a unique identifier but isn't called objectID,\n";
+            $message .= "you can map it automatically using `saveObjects(\$objects, ['objectIDKey' => 'primary'])`\n\n";
+            $message .= "Algolia is also able to generate objectIDs automatically but *it's not recommended*.\n";
```
rayrutjes: I feel like this raises more questions than it answers.
Why isn't it recommended?
Is the next sentence a good practice or a bad one?

src/Support/Helpers.php
```diff
@@ -118,4 +118,15 @@ public static function json_decode($json, $assoc = false, $depth = 512)
 
         return $data;
     }
+
+    public static function mapObjectIDs($objectIDKey, $objects)
+    {
+        return array_map(function ($object) use ($objectIDKey) {
+            if (!isset($object[$objectIDKey])) {
+                throw new MissingObjectId("One object didn't have any $objectIDKey key: ".json_encode($object));
```
rayrutjes: s/One object/At least one object/

rayrutjes: I'd also use the present tense.

`At least one object is missing the required "$objectIDKey": ...`

julienbourdeau: 👌

src/SearchIndex.php
julienbourdeau: Yes, I agree. I couldn't find better without going into too many details. It should only be used if you reindex everything at once and never update/delete records (like docsearch).

PR #479, opened 2018-11-07
Fixed listIndices name (was still listIndexes)

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | yes     
| Need Doc update   | no

## Describe your change

The method `listIndexes` was renamed `listIndices`, this is how it was documented: https://deploy-preview-2021--algolia-doc.netlify.com/doc/api-reference/api-methods/list-indices/

Unfortunately, I didn't change the name properly in the client.  😩 
I realized that looking at @vinkla's PR https://github.com/vinkla/laravel-algolia/pull/12

### DIFF
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -125,7 +125,7 @@ class SearchClient
         );
     }
 
-    public function listIndexes($requestOptions = array())
+    public function listIndices($requestOptions = array())
     {
         return $this->api->read('GET', api_path('/1/indexes/'), $requestOptions);
     }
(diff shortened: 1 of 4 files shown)


src/SearchClient.php
```diff
@@ -125,7 +125,7 @@ public function multipleGetObjects($requests, $requestOptions = array())
         );
     }
 
-    public function listIndexes($requestOptions = array())
```
nunomaduro: Shouldn't we update tests, upgrade guide and readme.

PR #480, opened 2018-11-07
Define public API better

julienbourdeau:
We added final on many classes and @internal tags on code
that shouldn't be used

| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | compare to RC, could be     
| Need Doc update   | yes

## Describe your change

Many classes are now final and some are tagged `@internal`. All `@internal` classes should be used, they could break at some point.

If you need to extend a class we marked as `final`, please let us know, there could be good reasons we missed.

### DIFF
--- a/src/RetryStrategy/HostCollection.php
+++ b/src/RetryStrategy/HostCollection.php
@@ -2,7 +2,10 @@
 
 namespace Algolia\AlgoliaSearch\RetryStrategy;
 
-class HostCollection
+/**
+ * @internal
+ */
+final class HostCollection
 {
     private $hosts;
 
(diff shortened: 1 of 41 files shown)


src/RetryStrategy/HostCollection.php
```diff
@@ -2,7 +2,10 @@
 
 namespace Algolia\AlgoliaSearch\RetryStrategy;
 
-class HostCollection
+/**
+ * @internal s
```
nunomaduro: Small typo here. The `s`.

src/SearchClient.php
```diff
@@ -6,16 +6,17 @@
 use Algolia\AlgoliaSearch\Response\MultipleIndexingResponse;
 use Algolia\AlgoliaSearch\Response\AddApiKeyResponse;
 use Algolia\AlgoliaSearch\Response\UpdateApiKeyResponse;
-use Algolia\AlgoliaSearch\RetryStrategy\ApiWrapper;
 use Algolia\AlgoliaSearch\RequestOptions\RequestOptions;
+use Algolia\AlgoliaSearch\RetryStrategy\ApiWrapper;
```
nunomaduro: You can remove this right?

PR #481, opened 2018-11-07
Move methods again 🤘😃

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | yes/no

## Describe your change

- [x] Introduce MultiResponse which is just a wrapper around more responses
- [x] Refactor copy*To, moveTo to be part of the client
- [x] Introduce `AccountClient::copyIndex($index1, $index2)`


src/AccountClient.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Algolia\AlgoliaSearch;
+
+use Algolia\AlgoliaSearch\Response\MultiResponse;
+
+class AccountClient
+{
+    public static function copyIndex(SearchIndex $srcIndex, SearchIndex $destIndex)
```
nunomaduro: What happens if the destination index already exists?

src/AccountClient.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace Algolia\AlgoliaSearch;
+
+use Algolia\AlgoliaSearch\Response\MultiResponse;
+
+class AccountClient
```
nunomaduro: Shouldn't this be final?

src/AccountClient.php
nunomaduro: There is a reason for making this `static` ?

src/Response/MultiResponse.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Response;
+
+class MultiResponse extends AbstractResponse implements \Iterator, \Countable
```
nunomaduro: Nice one! Can you add this response on methods that are returning more then one response? Example: `replaceAllObjects`?

src/AccountClient.php
julienbourdeau: it should

src/AccountClient.php
julienbourdeau: Currenly, Acccount doesnt have any state and doesnt have any dependencies. So I think static for now is good

src/Response/MultiResponse.php
julienbourdeau: Normally I did it everywhere
(replaceAllObjects: https://github.com/algolia/algoliasearch-client-php/pull/481/files#diff-8e5cca1a0496ef5410c39a80df3a03e8R230)

src/AccountClient.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace Algolia\AlgoliaSearch;
+
+use Algolia\AlgoliaSearch\Exceptions\NotFoundException;
+use Algolia\AlgoliaSearch\Response\MultiResponse;
+
+final class AccountClient
+{
+    public static function copyIndex(SearchIndex $srcIndex, SearchIndex $destIndex, $requestOptions = array())
```
Ant-hem: Correct me if I am wrong but `$requestOptions `are never used here?

julienbourdeau: Yes, never used yet but required for future-proof method signature. All methods, unless there is really a strong reason not to, have a RequestOptions last argument

src/AccountClient.php
julienbourdeau: Great catch! as discussed, the method will check if the destination exists and fail if it does.

src/AccountClient.php
```diff
@@ -0,0 +1,38 @@
...
```
nunomaduro: You need to check if the indexes are from different accounts.

julienbourdeau: I don't like to expose the appid from the Index but I see why we need to do that 👌

PR #482, opened 2018-11-13
Fixes issue while sending an empty batch

nunomaduro:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

The response `status` in a `POST` of an empty batch I will be always be `notPublished`, causing an issue while `wait()` this response.

### DIFF
--- a/src/SearchIndex.php
+++ b/src/SearchIndex.php
@@ -372,10 +372,12 @@ class SearchIndex
             }
         }
 
-        if ('addObject' !== $action) {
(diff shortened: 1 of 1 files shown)


julienbourdeau:
I can't really figure out if we should throw an error or not 🤔

nunomaduro:
@julienbourdeau We need to spec the `save_objects`, but I think we should return an NullResponseObject on those causes.

julienbourdeau:
Same goes for `save_rules` and `save_Synonyms`

PR #485, opened 2018-11-19
Fix json_encode error detection

jimbojsb:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no   
| BC breaks?        | no     
| Related Issue     | 
| Need Doc update   |no

## Describe your change

refactor detection of `json_encode()` errors

As currently written, with an empty body, the client will choose not to json_encode the body. However, if there is a previous `json_last_errror()` from any other json_encode calls during the request lifetime, the client will throw an exception, even though the error is not from _its own_ `json_encode()` call

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ composer.lock
 /vendor/
 .php_cs.cache
 /tests/cache/
+.idea/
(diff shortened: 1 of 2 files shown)


julienbourdeau:
Thanks a lot for your contribution!

PR #486, opened 2018-11-22
Include the composer autoloader

fruitl00p:
| Q                 | A
| Bug fix?          | yes ... well, sort of... 
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Also include the composer autoloader if available.

In order to actually check if the GuzzleClient is available the autoloader should be included. 

As this library might be installed without composer (“¯\_(ツ)_/¯“) it can't be a `require`-call but `include` will work... (sort of)

### DIFF
--- a/bin/algolia-doctor
+++ b/bin/algolia-doctor
@@ -2,7 +2,12 @@
 <?php
 
 // This kind of sucks.. but we need the composer autoloader, in order to fully check for GuzzleHttp
-include __DIR__ . '/../../../autoload.php';
+if (file_exists(__DIR__ . '/../../../autoload.php')) {
+    include __DIR__ . '/../../../autoload.php';
+} else {
+    // falback to the local non-composer autoloader which seems to fail to check for Guzzle
+    include __DIR__ . '/../autoload.php';
+}
 
 $doc = new AlgoliaDoctor();
 $doc->checkPhpVersion();
--- a/bin/algolia-doctor
+++ b/bin/algolia-doctor
@@ -1,6 +1,9 @@
 #!/usr/bin/env php
 <?php
 
+// This kind of sucks.. but we need the composer autoloader, in order to fully check for GuzzleHttp
+include __DIR__ . '/../../../autoload.php';
+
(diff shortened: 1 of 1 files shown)


bin/algolia-doctor
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env php
 <?php
 
+// This kind of sucks.. but we need the composer autoloader, in order to fully check for GuzzleHttp
+include __DIR__ . '/../../../autoload.php';
```
nunomaduro: We should be careful here. There are conditions, such as local development, where this file may not exists.

nunomaduro: I solution may be, if file exists, we include. @julienbourdeau What do you think?

julienbourdeau: Yes adding `if (file_exists(...))` before would be best. Include works indeed but throws a warning :(

fruitl00p: I've updated the PR to address this issue...

bin/algolia-doctor
```diff
@@ -1,6 +1,14 @@
 #!/usr/bin/env php
 <?php
 
+// This kind of sucks.. but we need the composer autoloader, in order to fully check for GuzzleHttp
+if (file_exists(__DIR__ . '/../../../autoload.php')) {
+    include __DIR__ . '/../../../autoload.php';
+} else {
+    // falback to the local non-composer autoloader which seems to fail to check for Guzzle
+    include __DIR__ . '/../autoload.php';
```
nunomaduro: I think we should also apply a `file_exists` here.

nunomaduro:
@TomKlotzPro Can you take this one?

TomKlotzPro:
I'm on it.

TomKlotzPro:
@fruitl00p we've corrected it thank you.

TomKlotzPro:
This issue has been fixed in the pull request: https://github.com/algolia/algoliasearch-client-php/pull/534.

PR #493, opened 2018-12-12
Introduce Get/Set Strategy for Personalization

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change

Introducing

```php
$strategy = $searchClient->getStrategy();
$searchClient->setStrategy($strategy);
```

### DIFF
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -242,6 +242,23 @@ class SearchClient
         return base64_encode($content);
     }
 
+    public function getPersonalizationStrategy($requestOptions = array())
+    {
+        return $this->api->read('GET', api_path('/1/recommendation/personalization/strategy'), $requestOptions);
+    }
+
+    public function setPersonalizationStrategy($strategy, $requestOptions = array())
+    {
+        $apiResponse = $this->api->write(
+            'POST',
+            api_path('1/recommendation/personalization/strategy'),
+            $strategy,
+            $requestOptions
+        );
+
+        return $apiResponse;
+    }
+
     public function searchUserIds($query, $requestOptions = array())
     {
         if (is_array($requestOptions)) {


src/SearchClient.php
```diff
@@ -242,6 +242,23 @@ public static function generateSecuredApiKey($parentApiKey, $restrictions)
...
+    public function setStrategy($strategy, $requestOptions = array())
+    {
+        $apiResponse = $this->api->write(
+            'POST',
+            api_path('1/recommendation/personalization/strategy'),
+            $strategy,
+            $requestOptions
+        );
+
+        return $apiResponse; // TODO: Create WaitableResponse once we know
```
nunomaduro: Yeah, as you specified missing response object here.

src/SearchClient.php
```diff
@@ -242,6 +242,23 @@ public static function generateSecuredApiKey($parentApiKey, $restrictions)
         return base64_encode($content);
     }
 
+    public function getStrategy($requestOptions = array())
+    {
+        return $this->api->read('GET', api_path('/1/recommendation/personalization/strategy'), $requestOptions);
+    }
+
+    public function setStrategy($strategy, $requestOptions = array())
```
nunomaduro: Missing test for this.

src/SearchClient.php
```diff
@@ -242,6 +242,23 @@ public static function generateSecuredApiKey($parentApiKey, $restrictions)
         return base64_encode($content);
     }
 
+    public function getStrategy($requestOptions = array())
```
nunomaduro: Missing test for this.

src/SearchClient.php
```diff
@@ -242,6 +242,23 @@ public static function generateSecuredApiKey($parentApiKey, $restrictions)
...
+    {
+        return $this->api->read('GET', api_path('/1/recommendation/personalization/strategy'), $requestOptions);
+    }
+
+    public function setStrategy($strategy, $requestOptions = array())
+    {
+        $apiResponse = $this->api->write(
+            'POST',
+            api_path('1/recommendation/personalization/strategy'),
+            $strategy,
```
tkrugg: The name of the function `getStrategy`/`setStrategy` suggests that they are generic.
Therefore shouldn't `personalization` be a parameter?

julienbourdeau: So you'd rather see `setPersonalizationStrategy`?

tkrugg: I think so, but I don't know if there is going to be any more use cases for strategies

src/SearchClient.php
nunomaduro: This should be, as discussed together: `getPersonalizationStrategy`.

src/SearchClient.php
nunomaduro: This should be, as discussed together: `setPersonalizationStrategy`.

src/SearchClient.php
nunomaduro: Yeah, as you specified missing response object here.

PR #499, opened 2018-12-21
[2.1.2] Adds `null` resilience to search $query

nunomaduro:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no
| Need Doc update   | no

## Describe your change

PHP is not a strong typed language, and some developers may enter in the mistake of passing non strings into a `search` method. E.g: In the Laravel Framework:

```php
$query = request('query'); // This may be null.

$results = Post::search($query);
```

This Pull Request addresses this cases, the `null` type is now converted to an empty string.

### DIFF
--- a/tests/Unit/RequestTestCase.php
+++ b/tests/Unit/RequestTestCase.php
@@ -29,7 +29,7 @@ abstract class RequestTestCase extends TestCase
     protected function assertBodySubset($subset, RequestInterface $request)
     {
         $body = json_decode((string) $request->getBody(), true);
-        $this->assertArraySubset($subset, $body);
+        $this->assertArraySubset($subset, $body, true);
     }
 
     protected function assertQueryParametersSubset(array $subset, RequestInterface $request)
--- a/tests/Unit/SearchTest.php
+++ b/tests/Unit/SearchTest.php
@@ -1,0 +1,55 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Tests\Unit;
+
+use Algolia\AlgoliaSearch\Exceptions\RequestException;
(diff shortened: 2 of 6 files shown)


nunomaduro:
@julienbourdeau You are right! Just added some tests on this.

tests/Unit/RequestTestCase.php
```diff
@@ -29,7 +29,7 @@ protected function assertEndpointEquals(RequestInterface $request, $endpoint)
     protected function assertBodySubset($subset, RequestInterface $request)
     {
         $body = json_decode((string) $request->getBody(), true);
-        $this->assertArraySubset($subset, $body);
+        $this->assertArraySubset($subset, $body, true);
```
nunomaduro: @julienbourdeau Nice improvement here, where this comparision is now `strict`.

tests/Unit/SearchTest.php
```diff
@@ -0,0 +1,55 @@
...
+        } catch (RequestException $e) {
+            $this->assertBodySubset(array('query' => '',), $e->getRequest());
+        }
+
+        $client = PlacesClient::create('id', 'key');
+
+        try {
+            $client->search(null);
+        } catch (RequestException $e) {
+            $this->assertBodySubset(array('query' => '',), $e->getRequest());
```
rayrutjes: `array('query' => '',)` that is a weird syntax fomatting, expected?

nunomaduro: @rayrutjes We are using the `symfony` cs preset, that rule seems to be on it.

PR #506, opened 2019-01-15
Fix autoloader when not using composer

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Working on WordPress integration, I realized that in some cases, the required files couldn't be found. I added `__DIR__` to make sure we always require the correct path.


nunomaduro:
Thanks for this.

PR #507, opened 2019-01-15
Remove configuration files from exported zip

julienbourdeau:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Remove some files from the zip generated by github. It should also be removed by composer when installing the lib.

These files are necessary for dev but shouldn't be distributed.

screenshot 2019-01-15 at 12 39 04

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,10 +2,13 @@
 
 /docs              export-ignore
 /tests             export-ignore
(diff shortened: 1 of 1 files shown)


nunomaduro:
Thanks for this!

PR #515, opened 2019-02-14
Fix iterator key increment

### LINKED ISSUE #514: browseObjects() Iterator keys not being set properly
- Algolia Client Version: 2.2.0
- Language Version: PHP 7.1

### Description
When using the `Algolia\AlgoliaSearch\SearchIndex@browseObjects` method, the Iterator that is returned doesn't increment the keys properly. I think you need to call `$this->key++` at the top of `Algolia\AlgoliaSearch\Iterators\AbstractAlgoliaIterator@next`

### Steps To Reproduce
Try to use `iterator_to_array()` on the Iterator that is returned. I know you can get round this by passing a second `false` parameter to the method, but it feels like an oversight since you've got a `$this->key` property on your Iterator.

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | Fix #514   
| Need Doc update   | no

## Describe your change

when browsing an index (objects, rules or synonyms), the "key" (index) of the loop was never incremented.

```php
$iter = $client->initIndex('something')->browseObjects();

foreach ($iter as $key => $value) {
    // without this fix, $key will always be 0
    dump($key);
}
```

### DIFF
--- a/src/Iterators/AbstractAlgoliaIterator.php
+++ b/src/Iterators/AbstractAlgoliaIterator.php
@@ -82,6 +82,7 @@ abstract class AbstractAlgoliaIterator implements \Iterator
      */
     public function next()
     {
+        $this->key++;
         $this->batchKey++;
         if ($this->valid()) {
             return;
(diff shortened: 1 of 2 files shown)


bgurney:
Any plans to merge this?

bgurney:
Thanks! When is this likely to be tagged?

nunomaduro:
Just wanting for tests to pass on travis.

nunomaduro:
@bgurney Version `2.2.1` has been released.

PR #523, opened 2019-03-04
Fix sffv function name (plural)

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

I made a mistake when naming the `searchForFacetValues` method in the v2. I forgot the S at the end. The documentation is "correct" (ie use plural method name).

This PR fix the name of the method and is aliasing the wrong name to the correct name.

### DIFF
--- a/src/SearchIndex.php
+++ b/src/SearchIndex.php
@@ -59,7 +59,15 @@ class SearchIndex
         return $this->api->read('POST', api_path('/1/indexes/%s/query', $this->indexName), $requestOptions);
     }
 
+    /**
+     * @deprecated Please use searchForFacetValues instead
+     */
     public function searchForFacetValue($facetName, $facetQuery, $requestOptions = array())
+    {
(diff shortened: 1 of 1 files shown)


src/SearchIndex.php
```diff
@@ -59,7 +59,15 @@ public function search($query, $requestOptions = array())
         return $this->api->read('POST', api_path('/1/indexes/%s/query', $this->indexName), $requestOptions);
     }
 
+    /**
+     * @deprecated Please use searchForFacetValues instead
+     */
     public function searchForFacetValue($facetName, $facetQuery, $requestOptions = array())
+    {
+        return searchForFacetValues($facetName, $facetQuery, $requestOptions);
```
nunomaduro: Tests are passing buddy? You forgot a $this here.

PR #526, opened 2019-03-08
[Discussion] #525: Add support for Symfony HTTP Client

### LINKED ISSUE #525: Symfony HttpClient adapter
Symfony release its new HttpClient: https://github.com/symfony/http-client/

I'm wondering if it would make sense to have an adapter for it, next to the PHP5.3 and guzzle ones.

As far as I understood, it's not PSR compliant so it might require a bit more work, to convert [requests and responses](https://github.com/algolia/algoliasearch-client-php/tree/master/src/Http).

The main reason I see to add it, is *if* it brings value to our [SearchBundle](htts://github.com/algolia/search-bundle).

kiler129:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #525
| Need Doc update   | yes

## Describe your change
Experimental / WIP implementation of the http://github.com/symfony/http-client

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 # Release Notes
 
 ## [Unreleased]
+- Add experimental support for Symfony HTTP Client ([#525](https://github.com/algolia/algoliasearch-client-php/issues/525))
 
 ## [v2.2.3](https://github.com/algolia/algoliasearch-client-php/compare/2.2.2...2.2.3)
 
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         }
     },
     "suggest": {
-        "guzzlehttp/guzzle": "If you prefer to use Guzzle HTTP client instead of the Http Client implementation provided by the package"
+        "guzzlehttp/guzzle": "If you prefer to use Guzzle HTTP client instead of the Http Client implementation provided by the package",
+        "symfony/http-client": "If you prefer to use Symfony HTTP client instead of the built-in Http Client implementation"
     },
     "bin": [
         "bin/algolia-doctor"
--- a/src/Algolia.php
+++ b/src/Algolia.php
@@ -4,6 +4,7 @@ namespace Algolia\AlgoliaSearch;
 
 use Algolia\AlgoliaSearch\Cache\NullCacheDriver;
 use Algolia\AlgoliaSearch\Http\HttpClientInterface;
+use Algolia\AlgoliaSearch\Http\SymfonyHttpClient;
 use Algolia\AlgoliaSearch\Log\DebugLogger;
 use Psr\Log\LoggerInterface;
 use Psr\SimpleCache\CacheInterface;
@@ -91,7 +92,9 @@ final class Algolia
     public static function getHttpClient()
     {
         if (null === self::$httpClient) {
-            if (class_exists('\GuzzleHttp\Client')) {
+            if (\class_exists('\Symfony\Component\HttpClient\HttpClient')) {
+                self::setHttpClient(new SymfonyHttpClient());
+            } elseif (class_exists('\GuzzleHttp\Client')) {
                 self::setHttpClient(new \Algolia\AlgoliaSearch\Http\Guzzle6HttpClient());
             } else {
                 self::setHttpClient(new \Algolia\AlgoliaSearch\Http\Php53HttpClient());
--- a/src/Http/SymfonyHttpClient.php
+++ b/src/Http/SymfonyHttpClient.php
@@ -1,0 +1,71 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Http;
+
+use Algolia\AlgoliaSearch\Exceptions\RetriableException;
+use Algolia\AlgoliaSearch\Http\Psr7\Response;
+use Psr\Http\Message\RequestInterface;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface;
+
+/**
+ * Implements experimental integration with Symfony HTTP Client.
+ *
+ * @see https://github.com/symfony/http-client/
+ * @experimental
+ */
+final class SymfonyHttpClient implements HttpClientInterface
+{
+    /**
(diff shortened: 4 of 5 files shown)


src/Algolia.php
```diff
@@ -91,7 +92,9 @@ public static function setLogger(LoggerInterface $logger)
     public static function getHttpClient()
     {
         if (null === self::$httpClient) {
-            if (class_exists('\GuzzleHttp\Client')) {
+            if (\class_exists('\Symfony\Component\HttpClient\HttpClient')) {
```
kiler129: Small comment regarding this - it takes a priority over Guzzle, since there's a high chance that someone has Guzzle in their project (and thus making Symfony HTTP Client impossible to be used).

src/Http/SymfonyHttpClient.php
```diff
@@ -0,0 +1,71 @@
...
+
+    public function sendRequest(RequestInterface $request, $timeout, $connectTimeout)
+    {
+        $opts = array(
+            'headers' => $request->getHeaders(),
+            'body' => (string) $request->getBody(),
+            'http_version' => '1.0' === $request->getProtocolVersion() ? '1.0' : null,
+        );
+
+        //In Symfony HTTP Client timeouts are abstracted and there's no separate timeouts, so pick higher one
```
kiler129: The question for reviewers: should we pick lowest or highest one?

julienbourdeau:
Thank you @kiler129 ! It looks very promising 👏 🎉 

I was expecting a lot more code. We'll review it next week and see how it work with our Symfony bundle.

src/Algolia.php
julienbourdeau: I'm thinking we could leave only Guzzle or Default here and ask people who want to use Symfony client to set it explicitly.
I think Guzzle should always be the default one. I see it as the "community's favorite" today.

Ideally, I'd like to add support for many clients but I'd rather not change the `getHttpClient` to handle them all. This instanciation is only here to add some magic and don't ask the users to set their HTTP client, unless they want to do somehting special.

kiler129: I also think such magic is a bad way of doing this, but the static/global `getHttpClient()` is called in multiple places.

For a Symfony Bundle / Laravel package we can preconfigure the client with a factory and simply allow setting a service in the configuration, so that anyone can decide what they want to use (with a default of Guzzle as the safest choice).

I don't know how will that apply in the scenario with SDK/client usage alone? Technically the setter can be called in the bootstrap of any application so maybe just a section in the docs describing how to use alternative HTTP Client?

kiler129:
Yes, I was expecting more code too or some problems along the line, but in fact it was very easy to integrate ;)

The Symfony bundle needs an update to support v2 (since now it's stuck on v1) since it's far superior. If you need any help I'm open.

nunomaduro:
Hey @kiler129, this pull request it's awesome.

I will try to find time during the week to check this out! 🥳

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 # Release Notes
 
 ## [Unreleased]
+- Add experimental support for Symfony HTTP Client ([#525](https://github.com/algolia/algoliasearch-client-php/issues/525))
```
nunomaduro:
```suggestion
- Add support for Symfony HTTP Client ([#525](https://github.com/algolia/algoliasearch-client-php/issues/525))
```

composer.json
```diff
@@ -38,7 +38,8 @@
         }
     },
     "suggest": {
-        "guzzlehttp/guzzle": "If you prefer to use Guzzle HTTP client instead of the Http Client implementation provided by the package"
+        "guzzlehttp/guzzle": "If you prefer to use Guzzle HTTP client instead of the Http Client implementation provided by the package",
+        "symfony/http-client": "If you prefer to use Symfony HTTP client instead of the built-in Http Client implementation"
```
nunomaduro:
```suggestion
        "symfony/http-client": "If you prefer to use Symfony HTTP client instead of the Http Client implementation provided by the package"
```

src/Http/SymfonyHttpClient.php
```diff
@@ -0,0 +1,71 @@
...
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface;
+
+/**
+ * Implements experimental integration with Symfony HTTP Client.
+ *
+ * @see https://github.com/symfony/http-client/
+ * @experimental
```
nunomaduro:
```suggestion
```

src/Algolia.php
nunomaduro: I am with @julienbourdeau on this one. Just remove the magic here, and we are going to set it `explicitly` on the `search_bundle`.

src/Http/SymfonyHttpClient.php
```diff
@@ -0,0 +1,71 @@
...
+use Algolia\AlgoliaSearch\Exceptions\RetriableException;
+use Algolia\AlgoliaSearch\Http\Psr7\Response;
+use Psr\Http\Message\RequestInterface;
+use Symfony\Component\HttpClient\HttpClient;
+use Symfony\Contracts\HttpClient\Exception\ClientExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ExceptionInterface;
+use Symfony\Contracts\HttpClient\Exception\ServerExceptionInterface;
+
+/**
+ * Implements experimental integration with Symfony HTTP Client.
```
nunomaduro:
```suggestion
```

src/Http/SymfonyHttpClient.php
```diff
@@ -0,0 +1,71 @@
...
+final class SymfonyHttpClient implements HttpClientInterface
+{
+    /**
+     * @var HttpClient
+     */
+    private $client;
+
+    public function __construct()
+    {
+        $this->client = HttpClient::create();
```
nunomaduro: Could you allow dependency injection here?

src/Http/SymfonyHttpClient.php
nunomaduro: Those are 2 different params.

nunomaduro:
This is a great proposal, but after some discussion I think this should not be on this low level client. Maybe in the future we will consider to add it to the search bundle.

THanks @kiler129 for this proposal, as is a great proposal and you have done a great pr! 👍

PR #535, opened 2019-03-22
Add Magento registration file

marcusirgens:
Allows the Magento implementation to generate interceptors.

| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   |no

## Describe your change
A file is added that checks if the library is running under Magento, and if so, it registers the component

Without this, it's clunky to create plugins and interceptors in Magento 2.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 # Release Notes
 
 ## [Unreleased]
+### Added
+- Magento registration file was added to allow interceptor creation
 
 ## [v2.2.3](https://github.com/algolia/algoliasearch-client-php/compare/2.2.2...2.2.3)
 
--- a/autoload.php
+++ b/autoload.php
@@ -18,6 +18,7 @@
  */
 require_once __DIR__.'/src/functions.php';
 require_once __DIR__.'/src/Http/Psr7/functions.php';
+require_once __DIR__.'/src/magento_registration.php';
 
 /*
  * Based on https://www.php-fig.org/psr/psr-4/examples/.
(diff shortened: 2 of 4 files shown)


julienbourdeau:
Ideally, this should be in the Magento extension, not this client. The client should be as agnostic as possible. I'm assuming it's possible to achieve the same in an Magento extension, otherwise it would be hard to use any lib with Magento but I'm not very familiar with the ecosystem.

Could you give us more details about why this has to be in the client and not in the Magento extension?

Thank you for your contribution 👍

marcusirgens:
@julienbourdeau If this is in the client, the client (as a separate composer package) has to either determine or assume where this package is located, which might cause issues if they are not installed using composer.

Here's a breakdown of it
```php
    \Magento\Framework\Component\ComponentRegistrar::register(
        \Magento\Framework\Component\ComponentRegistrar::LIBRARY, // Library (not a module)
        'algolia/algoliasearch-client-php', // library name - not very important
        __DIR__ // Registers the code in this directory and it's subdirectories
    );
```

[This is the relevant part of the Magento DI compilation code](https://github.com/magento/magento2/blob/677f1b6a463e4b1185ce33bb33d9cadf7c28db1a/setup/src/Magento/Setup/Console/Command/DiCompileCommand.php#L148-L151). Only registered code is targetted for interceptor generation.

I wholeheartedly agree that the best solution would be to put this in the Magento library, but I'm concerned that the logic for determining where this library is located in the filesystem will cause way more overhead than the extra `class_exists` call in cases where it's not installed under Magento.

nunomaduro:
Thanks @marcusirgens for your proposal.

Unfortunately we do not agree that this library should have code related to other libraries/frameworks.

PR #539, opened 2019-04-18
Adding count method to ObjectIterator

borealsmith:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change
Enables counting of the total number of objects in a given index by adding the count method to the OjectIterator. 

  

Determining the number of objects/records in an index


nunomaduro:
It's a great proposal, but at the moment there is no plans of adding this. Thanks!

PR #540, opened 2019-05-06
Custom install script escapes space in root path

julienbourdeau:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

If the user has a space in his root path like `/home/julien/My Projects`, the path will be broken in 2 arguments when passed to mkdir for instance.

This PR will escape space to turn the path into `/home/julien/My\ Projects`

### DIFF
--- a/bin/install-dependencies-without-composer
+++ b/bin/install-dependencies-without-composer
@@ -6,7 +6,7 @@
 //    true
 //);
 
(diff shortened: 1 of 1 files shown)


nunomaduro:
@julienbourdeau Going to release this once tests are passing, I am trying to figure out why they are not passing.

Issue #537, opened 2019-04-17
Guzzle 5 automatically wired but it's not supported


DelecroixQuentin:
- Algolia Client Version: 2.2.3
- PHP Version: 7.2

### Description
I am using algoliaserach-client-php with Prestashop, which includes Guzzle 5. The Algolia client finds and tries to use Guzzle, but it is integrated for Guzzle 6 ...

Except edit the Algolia client, is there a solution to choose the HTTP client to use rather than checking if Guzzle is installed (since the configuration maybe?)?

nunomaduro:
Hi @DelecroixQuentin,

That's a bug IMO. Here is the temporary workaround:

```php
$httpClient = new \Algolia\AlgoliaSearch\Http\Php53HttpClient();

\Algolia\AlgoliaSearch\Algolia::setHttpClient($httpClient);

$client = \Algolia\AlgoliaSearch\SearchClient::create('foo', 'bar');
```

DelecroixQuentin:
Hi @nunomaduro,

That works fine with your temporary workaround.
Thanks!

julienbourdeau:
Indeed we should probably check the guzzle version instead of checking for class_exists! 👍

TomKlotzPro:
I'm on it.

PR #544, opened 2019-05-10
Fix versions not supported for guzzle http client

TomKlotzPro:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no  
| BC breaks?        | no     
| Related Issue     | [#537](https://github.com/algolia/algoliasearch-client-php/issues/537)
| Need Doc update   | no

## Describe your change

This pull request is related to this issue [#537](https://github.com/algolia/algoliasearch-client-php/issues/537)

The PHP client has by default implementation for http client. 
However PHP client support guzzle as http client as well. 

A customer can use guzzle but only version 6. If he use a previous version it's not supported. 

In our current implementation, the PHP Client only check if the class exist. If the version is not supported it will automatically wired. Which result to a bug. 

Since the version 4, guzzle use in their 'ClientInterfact' the constant VERSION. I propose to use it to check the version.

### DIFF
--- a/src/Algolia.php
+++ b/src/Algolia.php
@@ -91,7 +91,7 @@ final class Algolia
     public static function getHttpClient()
     {
         if (null === self::$httpClient) {
-            if (class_exists('\GuzzleHttp\Client') && (\GuzzleHttp\Client::VERSION >= 6 && \GuzzleHttp\Client::VERSION < 7)) {
+            if (class_exists('\GuzzleHttp\Client') && (int) \GuzzleHttp\Client::VERSION[0] === 6) {
                 self::setHttpClient(new \Algolia\AlgoliaSearch\Http\Guzzle6HttpClient());
             } else {
                 self::setHttpClient(new \Algolia\AlgoliaSearch\Http\Php53HttpClient());
--- a/src/Algolia.php
+++ b/src/Algolia.php
@@ -37,7 +37,7 @@ final class Algolia
             return false;
         }
 
-        return !self::getCache() instanceof NullCacheDriver;
+        return ! self::getCache() instanceof NullCacheDriver;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Algolia.php
```diff
@@ -91,7 +91,7 @@ public static function setLogger(LoggerInterface $logger)
     public static function getHttpClient()
     {
         if (null === self::$httpClient) {
-            if (class_exists('\GuzzleHttp\Client')) {
+            if (\GuzzleHttp\Client::VERSION >= 6) {
```
nunomaduro: Hey buddy, two issues here:

- You need to make sure that the class `\GuzzleHttp\Client` exists
- You need to make sure that we are compatibly with Guzzle `^6.0` **only**.

TomKlotzPro: what should happen if the guzzleClient and the version is not correct ?

TomKlotzPro:
```php
if (class_exists('\GuzzleHttp\Client') && (\GuzzleHttp\Client::VERSION % 6) < 1)
```

TomKlotzPro:
For the moment, i check the version with `VERSION` as constant variable from [`InterfaceClient`](https://github.com/guzzle/guzzle/blob/master/src/ClientInterface.php)  in [`GuzzleHttp`](https://github.com/guzzle/guzzle). But it was implemented from version 4. If a user is using previous version, it will never find this variable. And it will result to a bug.

TomKlotzPro:
```php
const GUZZLE_VERSION_MIN = 6;

const GUZZLE_VERSION_MAX = 7;
```
@nunomaduro

nunomaduro:
No need for constants.

src/Algolia.php
```diff
@@ -91,7 +91,7 @@ public static function setLogger(LoggerInterface $logger)
     public static function getHttpClient()
     {
         if (null === self::$httpClient) {
-            if (class_exists('\GuzzleHttp\Client')) {
+            if (class_exists('\GuzzleHttp\Client') && (\GuzzleHttp\Client::VERSION >= 6 && \GuzzleHttp\Client::VERSION < 7)) {
```
nunomaduro: It's better if we make this explicit here.

```php
if (class_exists('\GuzzleHttp\Client') && (int) \GuzzleHttp\Client::VERSION[0] === 6) {
```

What do you think?

TomKlotzPro: Sounds good

PR #581, opened 2019-09-04
feat(remaining-validity): add getSecuredApiKeyRemainingValidity method

### LINKED ISSUE #576: New method: GetSecuredApiKeyRemainingValidity
## Summary

When a secured API keys expire, search breaks. The current workaround is more complex
than it needs to be. 

The options for developers using Algolia require, at every place where a search
operation is performed, catching an operation error, identify it as a
"ValidUntil" error, then regenerate the API key, and perform the search
operation again.

The idea is to define a method which tells if a `SecuredAPIKey`
is expired or not without making a single call to Algolia's API.

## References

- Specs: https://github.com/algolia/algoliasearch-client-specs-internal/pull/25

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #576
| Need Doc update   | yes

### DIFF
--- a/src/Exceptions/ValidUntilNotFoundException.php
+++ b/src/Exceptions/ValidUntilNotFoundException.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Exceptions;
+
+final class ValidUntilNotFoundException extends AlgoliaException
+{
+}
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -2,6 +2,7 @@
 
 namespace Algolia\AlgoliaSearch;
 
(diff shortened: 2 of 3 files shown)


src/Exceptions/ValidUntilNotFoundException.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Exceptions;
+
+final class ValidUntilNotFoundException extends BadRequestException
```
nunomaduro: This is not a bad request exception, as it doesn't call the api. You should extend the normal AlgoliaException one.

PR #584, opened 2019-09-09
fix(saveObjects-error-msg): refine the error thrown by the method

### LINKED ISSUE #583: Refine the error messages on saveObjects
When you pass a single array instead of an array of objects to the `saveObjects` method, the following error is returned: `PHP Fatal error:  Uncaught Algolia\AlgoliaSearch\Exceptions\MissingObjectId:
All objects must have an unique objectID (like a primary key) to be valid.`

The error should be changed to `PHP Fatal error: Please provide an array of objects instead of a single object.` to better reflect the problem. 

See https://secure.helpscout.net/conversation/950642494/173763/

chloelbn:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #583
| Need Doc update   | no

### DIFF
--- a/src/Exceptions/ArrayOfObjectsExpected.php
+++ b/src/Exceptions/ArrayOfObjectsExpected.php
@@ -1,0 +1,7 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Exceptions;
+
+class ArrayOfObjectsExpected extends AlgoliaException
+{
+}
--- a/src/SearchIndex.php
+++ b/src/SearchIndex.php
@@ -3,6 +3,7 @@
 namespace Algolia\AlgoliaSearch;
 
 use Algolia\AlgoliaSearch\Config\SearchConfig;
+use Algolia\AlgoliaSearch\Exceptions\ArrayOfObjectsExpected;
 use Algolia\AlgoliaSearch\Exceptions\MissingObjectId;
 use Algolia\AlgoliaSearch\Exceptions\NotFoundException;
 use Algolia\AlgoliaSearch\Exceptions\ObjectNotFoundException;
@@ -311,6 +312,10 @@ class SearchIndex
         $count = 0;
 
         foreach ($objects as $object) {
+            if (!is_array($object)) {
+                throw new ArrayOfObjectsExpected('Please provide an array of objects instead of a single object.');
+            }
+
             $batch[] = $object;
             $count++;
 
--- a/tests/Integration/SearchIndexTest.php
+++ b/tests/Integration/SearchIndexTest.php
@@ -6,6 +6,8 @@ use Algolia\AlgoliaSearch\SearchIndex;
 
 class SearchIndexTest extends AlgoliaIntegrationTestCase
 {
+    protected $index;
+
     protected function setUp()
     {
         parent::setUp();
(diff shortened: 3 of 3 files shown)


src/Exceptions/ArrayOfObjectsExpected.php
```diff
@@ -0,0 +1,7 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Exceptions;
+
+class ArrayOfObjectsExpected extends AlgoliaException
```
nunomaduro: `InvalidArgumentObjectsException`

src/SearchIndex.php
```diff
@@ -311,6 +312,10 @@ protected function splitIntoBatches($action, $objects, $requestOptions = array()
         $count = 0;
 
         foreach ($objects as $object) {
+            if (!is_array($object)) {
```
nunomaduro: You need to check if is Foreachable and not is array.

src/Exceptions/InvalidArgumentObjectsException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Exceptions;
+
+final class InvalidArgumentObjectsException extends AlgoliaException
+{
+    public function __construct($message = '', $code = 0, $previous = null)
+    {
+        if (!$message) {
```
nunomaduro: This is type juggling. Please use:

```suggestion
        if ($message === '') {
```

src/Exceptions/InvalidArgumentObjectsException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Exceptions;
+
+final class InvalidArgumentObjectsException extends AlgoliaException
+{
+    public function __construct($message = '', $code = 0, $previous = null)
```
nunomaduro: Missing public API.

src/Support/Helpers.php
```diff
@@ -130,4 +130,16 @@ public static function mapObjectIDs($objectIDKey, $objects)
...
+    /**
+     * Checks if a variable can be iterated through.
+     *
+     * @param mixed $object
+     *
+     * @return bool
+     */
+    public static function isIterable($object)
+    {
+        return is_array($object) || is_object($object) || $object instanceof \Traversable;
```
nunomaduro: Can you explain me `is_object($object)`?

chloelbn: I wanted to cover the maximum cases where a given variable is foreachable, and objects as arrays don't implement the `Traversable` interface. cf https://www.php.net/manual/en/class.traversable.php

nunomaduro: Yes, but `objects` are foreachable if they implement `ArrayAccess` interface. Correct? They `ArrayAccess` interface return `true` on `is_array`. No ?

src/Exceptions/InvalidArgumentObjectsException.php
```diff
@@ -0,0 +1,24 @@
...
+namespace Algolia\AlgoliaSearch\Exceptions;
+
+use Throwable;
+
+final class InvalidArgumentObjectsException extends AlgoliaException
+{
+    /**
+     * InvalidArgumentObjectsException constructor.
+     *
+     * @param string    $message  [optional] The Exception message to throw
```
nunomaduro: I would remove the `[optional]`. Is not idiomatic in PHP.

nunomaduro:
Putting this on hold, waiting for Chloe for this.

devdreamsolution (approved):

PR #599, opened 2019-11-21
feat: hasPendingMappings

### LINKED ISSUE #597: feat(mcm): new method has_pending_mapping
Please add the following method to `search_client`

```java
function has_pending_mappings(retrieve_mappings: boolean = false, opts: request_options) 
return has_pending_mappings_response

struct has_pending_mappings_response {
	pending: boolean,
	clusters: map<string, string[]>
}
```

```java
{
  "method": "GET",
  "url": "/1/clusters/mapping/pending?getClusters={retrieve_mappings}",
  "hostType": "read"
}
```

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #597
| Need Doc update   | yes

## Describe your change
- adds `hasPendingMappings` method

### DIFF
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -320,15 +320,15 @@ class SearchClient
     /**
      * Assign multiple userIds to the given cluster name.
      *
-     * @param  array<int, int> $userIds
-     * @param  string $clusterName    [description]
-     * @param  array  $requestOptions [description]
+     * @param array<int, int> $userIds
+     * @param string          $clusterName    [description]
+     * @param array           $requestOptions [description]
      *
      * @return array<string, mixed>
      */
     public function assignUserIds($userIds, $clusterName, $requestOptions = array())
-     {
-         return $this->api->write(
+    {
+        return $this->api->write(
              'POST',
              api_path('/1/clusters/mapping/batch'),
              array(
@@ -337,7 +337,7 @@ class SearchClient
              ),
              $requestOptions
          );
-     }
+    }
 
     public function removeUserId($userId, $requestOptions = array())
     {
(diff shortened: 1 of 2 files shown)


src/SearchClient.php
```diff
@@ -380,4 +380,21 @@ public static function getSecuredApiKeyRemainingValidity($securedAPIKey)
 
         return $validUntil - time();
     }
+
+    /**
+     * @param boolean $retrieveMappings
+     * @param array $requestOptions
+     * @return array<string, boolean|array>
```
nunomaduro: We can get an array out of this?

src/SearchClient.php
```diff
@@ -380,4 +380,21 @@ public static function getSecuredApiKeyRemainingValidity($securedAPIKey)
...
+     */
+    public function hasPendingMappings($retrieveMappings, $requestOptions = array())
```
nunomaduro: Tests are missing here, you are waiting for the CTS specs on this?

src/SearchClient.php
chloelbn: the response would look like 
```
[
  'pending' => true,
  'clusters' => [
    'foo-test' => ["bar-0000-00-00"]
  ]
]
```

src/SearchClient.php
```diff
@@ -380,4 +380,22 @@ public static function getSecuredApiKeyRemainingValidity($securedAPIKey)
...
         return $validUntil - time();
     }
+
+    /**
+     * @param bool  $retrieveMappings
+     * @param array $requestOptions
+     *
+     * @return array<string, boolean|array>
+     */
+    public function hasPendingMappings($retrieveMappings = false, $requestOptions = array())
```
julienbourdeau: I don't know if the name is already decided but usually when I see `hasSomething` method, I expect to get a boolean, not a "complex" object.
Maybe `getPendingMappings` could be a good name ¯\_(ツ)_/¯

chloelbn: it was indeed discussed that `getPendingMappings` would be a bit redundant with the parameter `retrieveMappings true|false`, and that `hasPendingMappings` would better convey what we're trying to achieve. Feel free to ping me if you want more information!

nunomaduro: As discussed, you need add `retrieveMappings` as part of request options. Just like : https://github.com/algolia/algoliasearch-client-python/pull/463/files.

PR #600, opened 2019-11-22
feat(recommendation-client): adds client and tests

### LINKED ISSUE #598: feat(API): introducing the Recommendation client
## Context

The personalization strategy endpoint will migrate from the Search API to the Recommendation API.

The consequences of the breaking changes are:

- In `search_client` both methods `get_personalization_strategy` and `set_personalization_strategy` **must** be deprecated.
- API Clients have to introduce a new `recommendation` client.

### Endpoint

`https://recommendation.{region}.algolia.com` where region can be `us` or `eu`.

Please note that as for the current `analytics_client` the region should be
`settable` when instantiating the client.

### Route

The route will be `/1/strategies/personalization` and support the following HTTP methods `GET`, `POST`, `OPTIONS`.

- `get_personalization_strategy` -> `GET` : `/1/strategies/personalization`
- `set_personalization_strategy` -> `POST` : `/1/strategies/personalization`

### recommendation_client interface

```java
interface recommendation_client {

    function get_personalization_strategy(opts: request_options)
    return get_strategy_response

    function set_personalization_strategy(set_strategy_request: strategy, opts: request_options)
    return set_strategy_response
}
```

### Objects

```java
struct set_strategy_request {
    eventsScoring: events_scoring[]
    facetsScoring: facets_scoring[]
    personalizationImpact: int
}
```

```java
struct events_scoring {
    eventName: string
    eventType: string
    score: int
}
```

```java
struct facets_scoring {
    facetName: string
    score: int
}
```

### Responses

```java
struct get_strategy_response {
    eventsScoring: events_scoring[]
    facetsScoring: facets_scoring[]
    personalizationImpact: int
}
```

```java
struct set_strategy_response {
    status: int
    message: string
}
```

For more details regarding the implementation please see the [specfication PR.](https://github.com/algolia/algoliasearch-client-specs-internal/pull/90)

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #598 
| Need Doc update   | yes

## Describe your change

* Deprecates legacy personalization endpoint
* adds new RecommendationClient


chloelbn:
Not completely sure how to test this, any suggestions welcomed

src/Config/RecommendationConfig.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Config;
+
+final class RecommendationConfig extends AbstractConfig
+{
+    public static function create($region, $appId = null, $apiKey = null)
```
nunomaduro: Can you add the public API here using PHP DOCS?

src/Config/RecommendationConfig.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        if (!$region || '' === $region) {
```
nunomaduro: I would use: `if (! is_string($region) || '' === $region) {`

src/Config/RecommendationConfig.php
```diff
@@ -0,0 +1,33 @@
...
+            throw new \InvalidArgumentException('The region is required');
+        }
+
+        $config = array(
+            'appId' => null !== $appId ? $appId : getenv('ALGOLIA_APP_ID'),
+            'apiKey' => null !== $apiKey ? $apiKey : getenv('ALGOLIA_API_KEY'),
+            'region' => $region,
+        );
+
+        return new static($config);
```
nunomaduro: This can be `new self`.

src/Config/RecommendationConfig.php
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    public function setRegion($region)
```
nunomaduro: What is the reason for this setter?

src/Config/RecommendationConfig.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        $this->config['region'] = $region;
+
+        return $this;
+    }
+
+    public function getRegion()
+    {
+        return $this->config['region'];
```
nunomaduro: What is the reason for this getter?

tests/Unit/RecommendationClientTest.php
```diff
@@ -0,0 +1,93 @@
...
+                array(
+                    'facetName' => 'brand',
+                    'score' => 10,
+                ),
+            ),
+            'personalizationImpact' => 75,
+        );
+
+        try {
+            self::$client->setPersonalizationStrategy($strategy);
```
nunomaduro: This is testable? I remember @aseure told me that we can't do it.

src/RecommendationClient.php
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * @param array<string, int|array>                       $strategy
+     * @param array<string, int|string|array>|RequestOptions $requestOptions
+     *
+     * @return array<string, int|string>
+     */
+    public function setPersonalizationStrategy($strategy, $requestOptions = array())
+    {
+        if (!array_key_exists('eventsScoring', $strategy)) {
+            throw new \InvalidArgumentException('eventsScoring cannot be null');
```
nunomaduro: I don't think we should validate this, the API already do it right?

src/RecommendationClient.php
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        if (!array_key_exists('facetsScoring', $strategy)) {
+            throw new \InvalidArgumentException('facetsScoring cannot be null');
```
nunomaduro: I don't think we should validate this, the API already do it right?

nunomaduro:
ps: don't forget to mark as deprecated the current methods.

tests/Unit/RecommendationClientTest.php
chloelbn: indeed it was just to test the payload, as I wad throwing errors if some parameters were missing at first. But now if I delete those checks I'm not sure what's left to test

src/Config/RecommendationConfig.php
chloelbn: it's needed to create the client with a config

nunomaduro:
@aseure Indeed, and you should have a created issue on go.

PR #610, opened 2020-02-07
feat(mcm): adds assignUserIDs method

chloelbn:
| Q                 | A
| ----------------- | ----------
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fix #586 
| Need Doc update   | yes

## Describe your change

Adds the MCM `assignUserIDs` method and adds the CTS MCM test suite

PR #611, opened 2020-02-07
fix(mcm): getTopUserId method

### LINKED ISSUE #609: Fix MCM `getTopUserId` method
MCM `getTopUserId` method does not use the right path. It needs to be fixed

chloelbn:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #609
| Need Doc update   | no

## Describe your change

There was an error in the path used for the `getTopUserId` method

### DIFF
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -296,7 +296,7 @@ class SearchClient
 
     public function getTopUserId($requestOptions = array())
     {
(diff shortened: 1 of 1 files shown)


nunomaduro:
Fixed in #610.

PR #621, opened 2020-04-20
Add PSR-18 HTTP client

### LINKED ISSUE #619: Missing support for Guzzle 7
Soon, Guzzle 7 will be out, and we need to add support to it: https://packagist.org/packages/guzzlehttp/guzzle

ostrolucky:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes
| BC breaks?        | no     
| Related Issue     | Fixes #619
| Need Doc update   | no

Currently there are only raw curl or Guzzle 6 adapters, while ecosystem is moving to abstractions. This can be seen with Guzzle 7 which fully implements PSR-18, or new Symfony HTTP client shipping PSR-18 adapter. Pretty much every PHP HTTP client takes care about PSR-18. We are ourselves not using Guzzle directly anymore, but through HttPlug. Hence we had to create such adapter ourselves in our project.

Moving forward, I would suggest to require PSR Http client in Algolia directly and deprecate current adapters, instead of having `Algolia\AlgoliaSearch\Http\HttpClientInterface`. Only difference it has is unnecessary $timeout and $connectTimeout parameters. In my opinion these should be configured inside wrapped http client, not overridden via your adapters.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,6 +23,7 @@
         "friendsofphp/php-cs-fixer": "^2.0",
         "fzaninotto/faker": "^1.8",
         "phpunit/phpunit": "^4.8",
+        "psr/http-client": "^1.0",
         "symfony/yaml": "^2.0 || ^4.0"
     },
     "autoload": {
--- a/src/Http/Psr18HttpClient.php
+++ b/src/Http/Psr18HttpClient.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Http;
+
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestInterface;
+
+class Psr18HttpClient implements HttpClientInterface
+{
+    /**
+     * @var ClientInterface
+     */
+    private $httpClient;
+
+    public function __construct(ClientInterface $httpClient)
+    {
+        $this->httpClient = $httpClient;
+    }
+
+    /**
+     * @inheritDoc
+     */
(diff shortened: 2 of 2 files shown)


ostrolucky:
Hmm so PSR-18 requires PHP 7.0. Ok to bump PHP requirement @nunomaduro ?

nunomaduro:
We are tackling issues ( including this one ) next quarter on the PHP Client. Do you mind waiting about 10 days to get a full answer on this? 🙏

ostrolucky:
@nunomaduro any news?

nunomaduro:
@ostrolucky I am sorry for keeping you waiting. We got other priorities at the moment, and I promise we will take a look into this in the next couple weeks.

chloelbn:
Hey @ostrolucky, unfortunately we have to keep the PHP requirement so dropping the interface is not possible. I'll take a look at your solution today!

ostrolucky:
Well this PSR interface requires PHP 7, so what would you want me to do so CI passes? Shall I drop requirement from composer.json?

chloelbn:
The tests fail at the moment because of the linter. Even though I agree your solution is more modern the constraints we have currently on the PHP versions we maintain prevent us from moving forward with it. Our current implementation of our Guzzle Client works with the version 7 so it's simpler if we keep it that way. I'm closing your PR but it's really cool you made this contribution, thanks!

ostrolucky:
In case somebody else is also interested in this, I have extracted this into composer package https://packagist.org/packages/psr18-adapter/algoliasearch-client-php

PR #628, opened 2020-06-02
fix: replace expired certificate within embedded certificate chain

aseure:
We previously embedded a full certificate chain for the sole purpose of
supporting environments where the system certificates are not available
or outdated.

However, on May 30th, 2020, one of the certificates within the
certificate chain we provided expired. This has the consequence of
making OpenSSL <1.1.1 and LibreSSL <3.2.0 consider the legit Algolia
certificate invalid, when it was not.

This commit replaces the faulty certificate from the embedded
certificate chain so that users who depends on it with outdated versions
of OpenSSL or LibreSSL can still prove Algolia servers identity.


nunomaduro:
Honestly I have no idea of the impact (or not) this could have.

May be worth to ask the customer to test with this branch:
```
composer require algolia/algoliasearch-client-php:"v1-replace-expired-certificate-dev"
```

aseure:
> Honestly I have no idea of the impact (or not) this could have.

I mean, it cannot be worse ^^

> May be worth to ask the customer to test with this branch

Yes for sure, thank you for providing the command by the way. I'm gonna write a draft answer.

aseure:
@chloelbn @asurak @nunomaduro 
We just got an answer from the customer:

>  Just an update. I tested import again and it works as it should. It seems it was a cache issue.

So technically, they don't even need the fix. I'm still gonna try it in isolation with up-to-date and outdated versions of the PHP client. If the fix is confirmed, I'm gonna prepare an FAQ entry for that cc @nunomaduro.

nunomaduro:
Ok ok.

The version `1.27` works in my side.

For testing locally with Symfony, we actually need this

```
composer require algolia/algoliasearch-client-php:"v1-replace-expired-certificate-dev as 1.27"
```

asurak:
> Ok ok.
> 
> The version `1.27` works in my side.
> 
> For testing locally with Symfony, we actually need this
> 
> ```
> composer require algolia/algoliasearch-client-php:"v1-replace-expired-certificate-dev as 1.27"
> ```

can you verify that the version before **doesn't** work?

nunomaduro:
> can you verify that the version before doesn't work?

The version `1.27` does work on my laptop. Also, @aseure just got a message from the customer that it also works on the customer side ( it was a cache issue they said ).

PR #633, opened 2020-07-10
Checks if apikey or appid exist in config array and throws exception …

florian25686:
…before contacting algolia

| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Check if required api settings (apikey and appid) are given in the config array or throw an exception.

If an api call is made with missing config information it retries multiple times and then fails with an exception. This generates not needed traffic and causing wait times for users.
This change handels the missing information before making changes and throwing an exception if the information are missing

### DIFF
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -56,7 +56,7 @@ class SearchClient
         $config = clone $config;
 
         if (!$config->getAppId() || !$config->getApiKey()) {
-            throw new UnreachableException('Missing config: AppId or AppKey. Please check your settings');
+            throw new UnreachableException('Missing config: appId or apiKey. Please check your settings');
         }
 
         $cacheKey = sprintf('%s-clusterHosts-%s', __CLASS__, $config->getAppId());
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -3,6 +3,7 @@
 namespace Algolia\AlgoliaSearch;
 
 use Algolia\AlgoliaSearch\Config\SearchConfig;
+use Algolia\AlgoliaSearch\Exceptions\UnreachableException;
 use Algolia\AlgoliaSearch\Exceptions\ValidUntilNotFoundException;
 use Algolia\AlgoliaSearch\RequestOptions\RequestOptions;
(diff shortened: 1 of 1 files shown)


nunomaduro:
How did you create an instance of the client so you could have that error? Can you send me the code snippet?

florian25686:
I followed the tutorial for laravel implementation and set the values in the .env-file. To test what happens if there is a problem with algolia, I set the `appId` to an empty string in my .env-file.

florian25686:
Hi @chloelbn,

who is maintaining this client in the future?
Any chance someone from algolia can follow up with this issue?

Best regards,
Florian

chloelbn:
Hi @florian25686,

I am one the maintainer of this client. We still have to review your PR but it will be done ASAP :) Thanks for your patience!

src/SearchClient.php
```diff
@@ -54,6 +55,10 @@ public static function createWithConfig(SearchConfig $config)
     {
         $config = clone $config;
 
+        if (!$config->getAppId() || !$config->getApiKey()) {
+            throw new UnreachableException('Missing config: AppId or AppKey. Please check your settings');
```
chloelbn, line 59: `Missing config: appId or apiKey. Please check your settings`

florian25686: Thank you for the input, I've made the requested change

src/SearchClient.php
```diff
@@ -54,6 +55,10 @@ public static function createWithConfig(SearchConfig $config)
     {
         $config = clone $config;
 
+        if (!$config->getAppId() || !$config->getApiKey()) {
+            throw new UnreachableException('Missing config: appId or apiKey. Please check your settings');
```
chloelbn, line 59: could you throw a `\InvalidArgumentException` here instead of `UnreachableException`?

chloelbn:
@florian25686 Could you add some tests and fix the ones that broke as well? Thanks a lot!

chloelbn (changes requested):

PR #637, opened 2020-08-04
Serialize query params in multipleQueries method

### LINKED ISSUE #624: Object forbidden at this position
We’re getting a bad request exception coming back from the API when including a `params` object in the query.

**Error**

```
Object forbidden at this position near line:1 column:56
```

**Code**

What our code looks like:

```php
$client = SearchClient::create($appId, $key);

$response = $client->multipleQueries([
	[
	    'indexName' => 'anyIndex',
	    'params' => [
	        'tagFilters' => ''
	    ],
	],
]);
```

**Request**

The request body (and the point that the exception references):

```json
{"requests":[{"indexName":"dev_opportunities","params":{"tagFilters":""}}]}
                                                       ↑
```

- If we remove the `params` object the request is successful
- According to the [docs](https://www.algolia.com/doc/api-reference/api-methods/multiple-queries/#method-param-query-object) `params` is valid
- The `params` is also present in the JavaScript client request that we’re trying to proxy (for security reasons)

---

- Algolia Client Version: 2.6.1
- Language Version: PHP

joshuabaker:
| Q                 | A
| Bug fix?          | no
| New feature?      | no 
| BC breaks?        | no     
| Related Issue     | Fix #624
| Need Doc update   | no

Please reference [the issue](https://github.com/algolia/algoliasearch-client-php/issues/624) for a description.

Maintainers, I suspect there may be a better home for the `serializeQueryParameters` method. Looking at the JavaScript client, [the like-for-like method](https://github.com/algolia/algoliasearch-client-javascript/blob/e6dec843932bdfaac13ffa1e0a11964975499078/packages/transporter/src/serializer.ts#L22-L36) is used in a few different places.

### DIFF
--- a/src/SearchClient.php
+++ b/src/SearchClient.php
@@ -15,6 +15,7 @@ use Algolia\AlgoliaSearch\RetryStrategy\ApiWrapper;
 use Algolia\AlgoliaSearch\RetryStrategy\ApiWrapperInterface;
 use Algolia\AlgoliaSearch\RetryStrategy\ClusterHosts;
 use Algolia\AlgoliaSearch\Support\Helpers;
+use const JSON_THROW_ON_ERROR;
 
 class SearchClient
 {
@@ -155,6 +156,12 @@ class SearchClient
 
     public function multipleQueries($queries, $requestOptions = array())
     {
+        $queries = array_map(function($query) {
+            $query['params'] = $this->serializeQueryParameters($query['params'] ?? []);
+
+            return $query;
+        }, $queries);
+
         if (is_array($requestOptions)) {
             $requestOptions['requests'] = $queries;
         } elseif ($requestOptions instanceof RequestOptions) {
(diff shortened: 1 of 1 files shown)


joshuabaker:
Guessing this is a version target getting missed?

Please review `serializeQueryParameters` and let me know your thoughts, and I can adjust for PHP compatibility.

chloelbn:
Hi @joshuabaker! Thanks for your contribution! Your fix looks good but before anything would you mind indeed fixing the PHP compatibility? We have to support PHP 5.6. Also would you mind adding some tests please?

As for where to put `serializeQueryParameters`, I'm okay if you put it in the functions.php file

src/SearchClient.php
```diff
@@ -155,6 +156,12 @@ public function isAlive($requestOptions = array())
 
     public function multipleQueries($queries, $requestOptions = array())
     {
+        $queries = array_map(function($query) {
```
chloelbn, line 159: Would you mind adding some tests, and fix the syntax for older PHP versions please?

chloelbn (changes requested):

chloelbn:
I'm closing this PR as it's been out of the scope for too long. If the feature is still needed please open a new one 🚀

PR #641, opened 2020-09-21
fix(personnalization-strategy): fixes flaky tests

### LINKED ISSUE #634: Implement retry until success for restore_api_key() call in CTS test
Because of the way Algolia is designed, performing a `restoreApiKey` may not
succeed because the previous `deleteApiKey` takes some time to get propagated
to all Algolia servers, giving us the following response: `[404] Key already
exists`.

To prevent that, we need to change the `testApiKeys` CTS test, if implemented,
to retry even on `[404] Key already exists` response.

chloelbn:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Fixes tests to skip the 429 error

### DIFF
--- a/tests/Integration/PersonalizationStrategyTest.php
+++ b/tests/Integration/PersonalizationStrategyTest.php
@@ -2,6 +2,7 @@
 
 namespace Algolia\AlgoliaSearch\Tests\Integration;
 
+use Algolia\AlgoliaSearch\Exceptions\BadRequestException;
 use Algolia\AlgoliaSearch\RecommendationClient;
 use PHPUnit\Framework\TestCase;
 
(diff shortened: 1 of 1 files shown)


DevinCodes (approved):

tests/Integration/PersonalizationStrategyTest.php
```diff
@@ -31,7 +31,13 @@ public function testPersonalizationStrategy()
             'personalizationImpact' => 0,
         );
 
-        $response = $recommendationClient->setPersonalizationStrategy($personalizationStrategy);
+        try {
+            $response = $recommendationClient->setPersonalizationStrategy($personalizationStrategy);
+        } catch (\Exception $e) {
```
DevinCodes, line 36: Can we catch an `AlgoliaException` here instead of the more generic `\Exception`?

DevinCodes (changes requested):

DevinCodes (approved):

DevinCodes:
Code looks good to me, so I approved ✅  
The CI is failing by the way, any idea why?

chloelbn:
@DevinCodes fixed the issue!

DevinCodes (approved):

PR #646, opened 2020-11-04
chore(ci): Migrate to Circle CI

julienbourdeau:
This PR moves the CI from Travis to CircleCI.

We're not going to test the client against PHP 5.3 anymore. Running such an old version of PHP is basically impossible today.

It runs against:

- [x] PHP 5.6
- [x] PHP 7.0
- [x] PHP 7.4
- [x] PHP _latest_ (7.4 today but it's going to be 8.0 very soon)


.circleci/config.yml
```diff
@@ -0,0 +1,90 @@
...
+              eval $(./algolia-keys export)
+              php vendor/bin/phpunit -v --testsuite Unit,Definition,Community
+            fi
+
+      - run:
+          name: Autoloading without Composer
+          command: php tests/tests-no-composer.php
+
+
+workflows:
```
chloelbn, line 80: Could you run the tests with Guzzle 6, 7 and the basic curl http requester please?

chloelbn (changes requested):

.circleci/config.yml
julienbourdeau: Damit I forgot about that ! Yes, thank you 👍

.circleci/config.yml
```diff
@@ -0,0 +1,90 @@
...
+      - run:
+          name: Run tests
+          command: |
+            export CI_BUILD_NUM=$CIRCLE_BUILD_NUM
+            if [ -z ${CIRCLE_PR_REPONAME+x} ]
+            then
+               php vendor/bin/phpunit
+            else
+              export CI_PROJ_USERNAME=$CIRCLE_PROJECT_USERNAME
+              export CI_PROJ_REPONAME=$CIRCLE_PROJECT_REPONAME
```
chloelbn, line 69: what are those two variables used for?

julienbourdeau: For the API Key Dealer client, in order to generate the correct payload

julienbourdeau:
PHP 8 is going to be released tomorrow.

I'm merging this to help testing on PHP 8. Since Travis is failing all over the place, it can only be better anyway.

Please let me know if there are any other change you'd like me to do.

PR #649, opened 2020-11-16
chore(docker): containerize the repo

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no 

## Describe your change

Add a `Dockerfile` to allow the use of the repo in a [Docker](https://www.docker.com/) container


DevinCodes (changes requested):
Thanks for the PR, @chloelbn ! Could you please apply the same changes as you did in https://github.com/algolia/algoliasearch-client-ruby/pull/437/files so they are consistent across the API clients? Thank you in advance! 🙌

DevinCodes (approved):

PR #650, opened 2020-11-17
refactor(synonyms): rename replaceExistingSynonyms parameter

### LINKED ISSUE #648: Rename `replaceExistingSynonyms` to `clearExistingSynonyms`
The parameter `replaceExistingSynonyms` in the `saveSynonyms` method is inconsistent with the rest of our API. We should rename it to `clearExistingSynonyms`. Backward-compatibility should be kept in mind

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #648 
| Need Doc update   | yes

## Describe your change

In the method `saveSynonyms`, rename the parameter `replaceExistingSynonyms` to `clearExistingSynonyms`

### DIFF
--- a/src/RequestOptions/RequestOptionsFactory.php
+++ b/src/RequestOptions/RequestOptionsFactory.php
@@ -12,7 +12,6 @@ final class RequestOptionsFactory
     private $validQueryParameters = array(
         'forwardToReplicas',
         'replaceExistingSynonyms',
-        'clearExistingSynonyms',
         'clearExistingRules',
         'getVersion',
     );
--- a/src/RequestOptions/RequestOptionsFactory.php
+++ b/src/RequestOptions/RequestOptionsFactory.php
@@ -12,6 +12,7 @@ final class RequestOptionsFactory
     private $validQueryParameters = array(
         'forwardToReplicas',
         'replaceExistingSynonyms',
+        'clearExistingSynonyms',
         'clearExistingRules',
         'getVersion',
     );
--- a/README.md
+++ b/README.md
@@ -62,3 +62,4 @@ Encountering an issue? Before reaching out to support, we recommend heading to o
 ## 📄 License
 
 Algolia PHP API Client is an open-sourced software licensed under the [MIT license](LICENSE.md).
+
--- a/src/RequestOptions/RequestOptions.php
+++ b/src/RequestOptions/RequestOptions.php
@@ -280,6 +280,13 @@ final class RequestOptions
         return $this;
     }
 
+    public function deleteBodyParameter($parameter)
+    {
+        unset($this->body[$parameter]);
+
+        return $this;
+    }
+
     /**
      * Replace all existing body parameters with the given name/value pairs.
      *
--- a/src/SearchIndex.php
+++ b/src/SearchIndex.php
@@ -379,6 +379,17 @@ class SearchIndex
 
     public function saveSynonyms($synonyms, $requestOptions = array())
     {
+        if (is_array($requestOptions)) {
+            if (array_key_exists('clearExistingSynonyms', $requestOptions)) {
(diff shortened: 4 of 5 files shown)


chloelbn:
@DevinCodes it could be argued that the fact it doesn't break the behaviour of the `replaceAllSynonyms` method in itself is enough to ensure the fix works. What do you think?

src/RequestOptions/RequestOptionsFactory.php
```diff
@@ -12,6 +12,7 @@ final class RequestOptionsFactory
     private $validQueryParameters = array(
         'forwardToReplicas',
         'replaceExistingSynonyms',
+        'clearExistingSynonyms',
```
Haroenv, line 15: doesn't this parameter need to be removed from the request? Now it's mapped to replaceExistingSynonyms, but it will still be sent, right?

chloelbn: what this list does it telling which parameter to handle as query param and build in the url, and which belongs to the body. but if the parameter is not set, it won't be sent. Does that answer your question?

chloelbn: it will be sent but will be ignored by the engine

DevinCodes: Good catch @Haroenv ! 

@chloelbn can we still remove it in that case? There's no need to send additional data to the engine if it doesn't use it anyway.

DevinCodes:
> @DevinCodes it could be argued that the fact it doesn't break the behaviour of the `replaceAllSynonyms` method in itself is enough to ensure the fix works. What do you think?

I think we should have a test with both parameters so that we're 100% sure there's backwards compatibility. WDYT?

src/RequestOptions/RequestOptionsFactory.php
chloelbn: @DevinCodes @Haroenv my reflexion on the moment was that as we're not checking what's sent to the engine in the dynamically typed languages clients, it would go in the same logic to send the parameter anyway as well. On the other side, deleting it would mean add a logic that doesn't exist in the `RequestOptions` class for the moment, because there is no method to do that.
A discussion could also be had with the engine to rename internally the parameter once all our PRs are released.

What are your thoughts folks?

Haroenv: I was actually mostly surprised that the engine ignores unknown parameters. I didn't know that.

Removing the parameter wouldn't really be hard by unsetting it on this line

https://github.com/algolia/algoliasearch-client-php/pull/650/files#diff-8894eed6b20bb44260a6d2ff1115052a7b8cacebeab972bb072195a7fc1f4357R384

and this line:

https://github.com/algolia/algoliasearch-client-php/pull/650/files#diff-8894eed6b20bb44260a6d2ff1115052a7b8cacebeab972bb072195a7fc1f4357R389

chloelbn: @Haroenv what I meant was that when the request options are not an array but a `RequestOptions` object, a method `deleteBodyParameter` would have to be added in that class in order to get rid of that param and I had doubts about it. But nevermind, I added it anyway while adding the test so we can make sure it's not sent. (I didn't see your comment before deleting mine, that's too bad)

chloelbn:
@DevinCodes @Haroenv tests added!

Haroenv (approved):

Haroenv:
test seems to fail based on setup though, is that expected?

Haroenv (approved):

chloelbn:
closing this for the moment, will reopen if the changes were to happen

PR #644, opened 2020-11-03
PHP 8 Support

driesvints:
| Q                 | A
| ----------------- | ----------
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Related Issue     | Fix #...  
| Need Doc update   | no

## Describe your change

This PR will provide PHP 8 Support for the library. Atm this is blocking https://github.com/laravel/scout/pull/425

I'll keep the PR in draft until I've fixed the build.

PR #652, opened 2020-12-10
PHP 8 Support with PHPUnit ^8.5 (9.3 wouldn't work)

thadbryson:
| Q                 | A
| ------------------------- | ----------
| Bug fix?                  | no
| New feature?         | yes
| BC breaks?            | no     
| Related Issue         | 
| Need Doc update   | no

## Describe your change

Provides PHP 8 Support with PHPUnit working up until ^9.3. I have not been able to test with PHP ^5.3. And I haven't been able to test the following:

- Algolia\AlgoliaSearch\Tests\Integration\MultiClusterManagementTest::testMultiClusterManagement
- Algolia\AlgoliaSearch\Tests\Integration\PersonalizationStrategyTest::testPersonalizationStrategy

I don't have those features on my Algolia account. 

I'm not sure if PHPUnit ^9.3 is better over 8.5. I may be able to change to 8.5 

PHP 8 support for this project is blocking PHP 8 support for [laravel/scout](https://github.com/laravel/scout) and Laravel Nova.
I would of edited the code for https://github.com/algolia/algoliasearch-client-php/pull/644 pull request but couldn't. I thank that author for the code I used.


thadbryson:
OK...so PHPUnit 9 did not work out. Going with 8.5. Will try that.

driesvints:
This won't work, not even with PHPUnit since the new void return type will fail on anything below PHP 7

thadbryson:
Yeah I'm working on using PHPUnit 8.5. I removed the void return types. I got it working except for Guzzle and the tests I mentioned where my Algolia account doesn't have access.

driesvints:
@thadbryson this will never work because Algolia needs support all the way from PHP 5.3 til PHP 8

thadbryson:
Yeah I just saw PHP-CS-Fixer doesn't support the 5.3 to 8 range. I bumped the version of it in my composer.json but can't test it on 5.3. I'll have to cancel this pull request. I agree with @driesvints  that they'll have to separate versions to support PHP 8.

greg0ire:
@driesvints @thadbryson possible solution for the `: void` issue: https://twitter.com/s_bergmann/status/1333483559181574148 ?

driesvints:
@greg0ire unfortunately that won't work either. This entire test suite is so old that there's much more issues than just the void return type. We simply can't use newer PHPUnit versions with the current test suite and the current PHPUnit v4 fails on PHP 8. Updating the entire test suite won't allow it to run on PHP <7.? anymore. We're in a stalemate where only a new major version with the dropping of older PHP versions will be a path forward.

I'm more than happy to be proven wrong though if anyone's up for a different solution.

thadbryson:
I think it'll need a v3 like @driesvints has suggested. Supporting PHP ^7.1 and PHPUnit ^9.3. I'm working on that now.

greg0ire:
> I'm more than happy to be proven wrong though if anyone's up for a different solution.

No, it was just in case `: void` was the only issue, but if there are a lot more issues then dropping support for PHP < 7.1 is probably the most sensible solution 👍

PR #665, opened 2021-02-04
Add Algolia CTS

damcou:
This PR changes the Test Suite and make it compliant with the Algolia CTS.


chloelbn (changes requested):

tests/Integration/InsightsClientTest.php
```diff
@@ -3,97 +3,149 @@
...
-        ), $response);
+        // clicked_object_ids
+        $insightUser = $insightsClient->user('bar');
+        $response = $insightUser->clickedObjectIDs(
+            'foo',
+            static::$indexes['sending_events'],
+            array('one', 'two')
+        );
+
+        self::assertEquals(200, $response['status']);
```
chloelbn, line 71: could you use the `$this->assertEquals` notation instead?

chloelbn (changes requested):

tests/Integration/BaseTest.php
```diff
@@ -0,0 +1,33 @@
...
+    {
+        try {
+            TestHelper::checkEnvironmentVariables();
+        } catch (\Exception $e) {
+            echo $e->getMessage()."\n";
+            exit(255);
+        }
+
+        parent::setUpBeforeClass();
+        static::$indexes = array();
```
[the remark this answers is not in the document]
damcou: It comes from the original class : https://github.com/algolia/algoliasearch-client-php/blob/master/tests/Integration/AlgoliaIntegrationTestCase.php#L22

To be honest, I don't fully know why it's there, I chose to keep it but I guess it's not mandatory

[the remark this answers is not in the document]
chloelbn: Yup, we can delete it! thanks

tests/Integration/AccountTest.php
```diff
@@ -0,0 +1,109 @@
...
+namespace Algolia\AlgoliaSearch\Tests\Integration;
+
+use Algolia\AlgoliaSearch\AccountClient;
+use Algolia\AlgoliaSearch\Response\MultiResponse;
+use Algolia\AlgoliaSearch\SearchIndex;
+use Algolia\AlgoliaSearch\Tests\TestHelper;
+
+class AccountTest extends BaseTest
+{
+    protected static $secondaryIndexes = array();
```
chloelbn, line 12: you can remove the `static` type now that you don't delete the indexes anymore

tests/Integration/AccountTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+    public function testCopyIndex()
+    {
+        static::$indexes['copy_index'] = TestHelper::getTestIndexName('copy_index');
```
chloelbn, line 16: same here, they can be called within the class

tests/Integration/AccountTest.php
```diff
@@ -0,0 +1,109 @@
...
+        static::$indexes['copy_index_2'] = TestHelper::getTestIndexName('copy_index_2');
+
+        /** @var SearchIndex $copyIndex */
+        $copyIndex = TestHelper::getClient()->initIndex(static::$indexes['copy_index']);
```
chloelbn, line 20: after making that change you'll be able to add the index directly in the indexes arrays and call `getIndexName()` to retrieve the name of each index anytime you'll need it

chloelbn (changes requested):

tests/Integration/RecommendationClientTest.php
```diff
@@ -0,0 +1,40 @@
...
+                array('eventName' => 'Purchase', 'eventType' => 'conversion', 'score' => 100),
+            ),
+            'facetsScoring' => array(
+                array('facetName' => 'brand', 'score' => 100),
+                array('facetName' => 'categories', 'score' => 10),
+            ),
+            'personalizationImpact' => 0,
+        );
+
+        $recommendationClient->setPersonalizationStrategy($strategy);
```
chloelbn, line 28: you can delete this part

chloelbn (changes requested):

tests/Integration/SearchClientTest.php
```diff
@@ -2,32 +2,445 @@
...
+        $this->assertArrayHasKey('pending', $response);
+    }
+
+    public function testApiKeys()
+    {
+        $acl = array('search');
+
+        $params = array(
+            'description' => 'A description',
+            'indexes' => array('index'),
```
chloelbn, line 243: maybe that'd be worth here to use a safe index name as so to avoid duplication with other test suite

chloelbn (changes requested):

tests/Integration/AccountTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+        try {
+            AccountClient::copyIndex($copyIndex, $secondaryIndex)->wait();
+        } catch (\Exception $e) {
+            $this->assertInstanceOf('\InvalidArgumentException', $e);
+            $this->assertEquals(
+                'Destination index already exists. Please delete it before copying index across applications.',
+                $e->getMessage()
+            );
+        }
```
DevinCodes, line 107: Same comment as before

tests/Integration/AnalyticsClientTest.php
```diff
@@ -3,50 +3,176 @@
...
+        $response = $analyticsClient->deleteABTest($abTestId);
+        $index->waitTask($response['taskID']);
+
+        try {
+            $result = $analyticsClient->getABTest($abTestId);
+        } catch (\Exception $e) {
+            $this->assertInstanceOf('Algolia\AlgoliaSearch\Exceptions\NotFoundException', $e);
+            $this->assertEquals(404, $e->getCode());
+            $this->assertEquals('ABTestID not found', $e->getMessage());
+        }
```
DevinCodes, line 110: Same comment as earlier

tests/Integration/AnalyticsClientTest.php
```diff
@@ -3,50 +3,176 @@
...
 
-        $response = $analyticsClient->deleteABTest($abTest['abTestID']);
-        $searchClient->waitTask(static::$indexes['aa_testing'], $response['taskID']);
+        try {
+            $result = $analyticsClient->getABTest($aaTestId);
+        } catch (\Exception $e) {
+            $this->assertInstanceOf('Algolia\AlgoliaSearch\Exceptions\NotFoundException', $e);
+            $this->assertEquals(404, $e->getCode());
+            $this->assertEquals('ABTestID not found', $e->getMessage());
+        }
```
DevinCodes, line 176: Same comment as earlier

tests/TestHelper.php
```diff
@@ -0,0 +1,116 @@
...
+
+    public static function formatRule($rule)
+    {
+        if (isset($rule['_metadata'])) {
+            unset($rule['_metadata']);
+        }
+
+        if (isset($rule['_highlightResult'])) {
+            unset($rule['_highlightResult']);
+        }
```
DevinCodes, line 112: We don't have to check if a key is set when using `unset`, and we can `unset` multiple keys in one call 🙂 
```suggestion
        unset($rule['_metadata'], $rule['_highlightResult']);
```

tests/Integration/SecuredApiKeysTest.php
```diff
@@ -0,0 +1,87 @@
...
+        $res = $securedIndex->search('');
+        $this->assertCount(1, $res['hits']);
+
+        try {
+            $res = $securedIndexDev->search('');
+        } catch (\Exception $e) {
+            $this->assertInstanceOf('Algolia\AlgoliaSearch\Exceptions\BadRequestException', $e);
+            $this->assertEquals(403, $e->getCode());
+            $this->assertEquals('Index not allowed with this API key', $e->getMessage());
+        }
```
DevinCodes, line 64: Same comment as earlier

tests/Integration/RecommendationClientTest.php
```diff
@@ -0,0 +1,40 @@
...
+
+        try {
+            $recommendationClient->setPersonalizationStrategy($strategy);
+        } catch (\Exception $e) {
+            $this->assertEquals(429, $e->getCode());
+        }
```
DevinCodes, line 32: Same comment as earlier

DevinCodes (dismissed):
Awesome job on this! I just have a few comments, but it looks great 💯

tests/Integration/AccountTest.php
```diff
@@ -0,0 +1,109 @@
...
+
+        try {
+            AccountClient::copyIndex($copyIndex, $secondaryIndex)->wait();
+        } catch (\Exception $e) {
+            $this->assertInstanceOf('\InvalidArgumentException', $e);
+            $this->assertEquals(
+                'If both index are on the same app, please use SearchClient::copyIndex method instead.',
+                $e->getMessage()
+            );
+        }
```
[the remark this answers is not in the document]
chloelbn: the thind is that this portion of code would then need to be extracted to its own function, whereas for the CTS we favour the approach of having all test in the same method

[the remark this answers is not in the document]
DevinCodes: Sorry, I mentioned the `expectsException` methods, but that's in another PHPUnit version. We can use the `setExpectedException` method. I can do the following in the same method without affecting the other assertions, so there doesn't seem to be a need to extract it into its own function:

```
$this->setExpectedException('\InvalidArgumentException', 'If both index are on the same app, please use SearchClient::copyIndex method instead.');
AccountClient::copyIndex($copyIndex, $secondaryIndex)->wait();
```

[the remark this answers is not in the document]
damcou: Hi @DevinCodes , from what I can see, doing things like this works but it stops the current test (so all the next tests afterwards are skipped), Am I doing something wrong ?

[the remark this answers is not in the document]
DevinCodes: Ah you're right, I'm seeing the same thing now indeed. My bad, let's keep the `try/catch` block then!

chloelbn (approved):

PR #667, opened 2021-02-12
fix: use `::class` instead of string for class name

matzeeable:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | no

## Describe your change

Use the `::class` identifier to get the full qualified name for a class instead of a string.

I want to bundle this package within my WordPress plugin together with [PHP-Scoper](https://github.com/humbug/php-scoper), unfortunately it does not scope a function within `functions.php` correctly.

> PHP-Scoper is a tool which essentially moves any body of code, including all dependencies such as vendor directories, to a new and distinct namespace.

PHP-Scoper does currently not support scoping array callbacks, see also https://github.com/humbug/php-scoper/issues/294


chloelbn:
Hi @matzeeable ! could you try and rebase on master please to see if this fixes the tests in the CI? Thanks

matzeeable:
Hey @chloelbn ! Just rebased from upstream, unfortunately the CI failed again. Can you please check what happens here? I think the code changes shouldn't be the problem. :-)

chloelbn:
Hi @matzeeable ! Indeed we need to solve this issue for the CI for ext contributors. I'll run your PR locally and merge it without the CI is everything goes fine

matzeeable:
Hey @chloelbn ! Just resolved the conflict :-)

PR #671, opened 2021-03-24
chore(cts): fix api keys test

### LINKED ISSUE #669: Fix flakiness for testApiKeys
The CI on some occasions may throw the following error `Algolia\AlgoliaSearch\Exceptions\NotFoundException: Key already exists`. We need to fix this flakiness

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #669

## Describe your change

Use a workaround to wait on api keys to be completely deleted when using restore

### DIFF
--- a/tests/Integration/SearchClientTest.php
+++ b/tests/Integration/SearchClientTest.php
@@ -303,11 +303,34 @@ class SearchClientTest extends BaseTest
             $this->assertInstanceOf('Algolia\AlgoliaSearch\Exceptions\NotFoundException', $e);
         }
 
-        TestHelper::getClient()->restoreApiKey($res['key'])->wait();
+        $retry = 1;
+        $time = 100000;
+        $maxRetries = 100;
+        do {
+            try {
+                TestHelper::getClient()->restoreApiKey($res['key'])->wait();
+            } catch (NotFoundException $e) {
+                $retry++;
+                $factor = ceil($retry / 10);
+                usleep($factor * $time); // 0.1 second
+                continue;
+            }
 
-        $restoredApiKey = TestHelper::getClient()->getApiKey($res['key']);
-        $this->assertEquals($acl, $restoredApiKey['acl']);
-        $this->assertEquals($params['description'], $restoredApiKey['description']);
+        } while ($retry >= $maxRetries);
+
+        do {
+            try {
+                $restoredApiKey = TestHelper::getClient()->getApiKey($res['key']);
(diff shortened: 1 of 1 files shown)


tests/Integration/SearchClientTest.php
```diff
@@ -303,11 +303,34 @@ public function testApiKeys()
             $this->assertInstanceOf('Algolia\AlgoliaSearch\Exceptions\NotFoundException', $e);
         }
 
-        TestHelper::getClient()->restoreApiKey($res['key'])->wait();
+        $retry = 1;
+        $time = 100000;
```
DevinCodes, line 307:
```suggestion
        $second = 100000;
```

tests/Integration/SearchClientTest.php
```diff
@@ -303,11 +303,34 @@ public function testApiKeys()
...
+        $maxRetries = 100;
+        do {
+            try {
+                TestHelper::getClient()->restoreApiKey($res['key'])->wait();
+            } catch (NotFoundException $e) {
+                $retry++;
+                $factor = ceil($retry / 10);
+                usleep($factor * $time); // 0.1 second
```
DevinCodes, line 315:
```suggestion
                usleep($factor * $second); // 0.1 second
```

tests/Integration/SearchClientTest.php
```diff
@@ -303,11 +303,34 @@ public function testApiKeys()
...
+
+        do {
+            try {
+                $restoredApiKey = TestHelper::getClient()->getApiKey($res['key']);
+                $this->assertEquals($acl, $restoredApiKey['acl']);
+                $this->assertEquals($params['description'], $restoredApiKey['description']);
+            } catch (NotFoundException $e) {
+                $retry++;
+                $factor = ceil($retry / 10);
+                usleep($factor * $time); // 0.1 second
```
DevinCodes, line 329:
```suggestion
                usleep($factor * $second); // 0.1 second
```

DevinCodes (changes requested):
Using `$second` as a name makes it clearer what the variable does IMO. It could even be a constant instead of a normal variable. WDYT?

TomKlotzPro:
Nice work @chloelbn 
Perhaps a refactor to avoid having duplicated code could be nice. 
Also I don't understand why we are doing this retry at some part of the code and not everywhere? 

+ I relaunch the CI and test are still failing 😭

Issue #668, opened 2021-02-16
Fix flakiness for AB/AAtests


chloelbn:
Sometimes the CI would throw `Algolia\AlgoliaSearch\Exceptions\NotFoundException: Index does not exist` on ABtesting and AAtesting. We need to fix this flakiness

PR #672, opened 2021-03-26
chore(CTS): add retry on tests

### LINKED ISSUE #669: Fix flakiness for testApiKeys

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no     
| Related Issue     | Fix #669 & #668
| Need Doc update   | no

## Describe your change
Adds a retry strategy to prevent flakiness


tests/Integration/AnalyticsClientTest.php
```diff
@@ -61,7 +61,10 @@ public function testAbTesting()
             $cpt++;
         } while (false);
 
-        $response = $analyticsClient->addABTest($abTest);
+        $response = TestHelper::retry(function () use ($analyticsClient, $abTest) {
+            return $analyticsClient->addABTest($abTest);
+        }, 1000, 40);
```
DevinCodes, line 66: Shouldn't this be `100000` everywhere since we use `usleep`?
```suggestion
        }, 100000, 40);
```

tests/TestHelper.php
```diff
@@ -107,4 +108,18 @@ public static function formatRule($rule)
 
         return $rule;
     }
+
+    public static function retry($function, $delay, $maxRetries)
```
DevinCodes, line 112: Since the `delay` and `maxRetries` are almost always the same in our test suite, what do you think of adding `100000` and `40` as the default values? This makes the code we use to retry more readable IMO, because we don't have to pass meaningless (at first sight) values like this. WDYT?
```suggestion
    public static function retry($function, $delay = 100000, $maxRetries = 40)
```

DevinCodes (changes requested):
This is a lot cleaner, it looks very nice! 😄  Let me know what you think of my comments.

DevinCodes (approved):

DevinCodes (changes requested):
It seems like the tests are failing

chloelbn:
@DevinCodes I'm doing this as we speak

tests/TestHelper.php
```diff
@@ -107,4 +108,18 @@ public static function formatRule($rule)
 
         return $rule;
     }
+
+    public static function retry($function, $delay = 100000, $maxRetries = 30)
```
DevinCodes, line 112: What would you say of changing the `delay` so that we can pass it in seconds (or milliseconds)? I think it'll make the code more readable, while keeping the choice to use `usleep` hidden to users. 

In that case the default parameter value would be `$delay = 0.1`, and in the catch block we would use `usleep($delay * $i * 1000000)` if we go for seconds, or `$delay = 1000` and `usleep($delay * $i * 100)` if we go for milliseconds.

WDYT?

chloelbn: I think it adds complexity on the contrary. I understand your idea but on the other side it makes an additional element within the sleep calculation in the catch block.

DevinCodes: Yes, but this is hidden to the user, which makes the API of this method easier to use. Right now, we need them to know they should use microseconds instead of the more conventional (milli)-seconds, which adds complexity for the user.

I agree we introduce a form of complexity, but I think the cost of adding it here outweigh the benefits of not doing so.

chloelbn: In that case I'd prefer using `sleep()` with `delay = 0.1` do have the same logic as in the Ruby client. WDYT?

DevinCodes (approved):

DevinCodes (changes requested):
I approved because the changes look good, but the CI is failing for (it seems) unrelated things. Could you quickly take a look please? 🙂  Thank you in advance!

PR #674, opened 2021-04-05
chore: updates phpunit

chloelbn:
| Q                 | A
| Bug fix?          | no
| New feature?      | no
| BC breaks?        | no

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,7 +86,7 @@ jobs:
               export CI_PROJ_USERNAME=$CIRCLE_PR_USERNAME
               export CI_PROJ_REPONAME=$CIRCLE_PR_REPONAME
 
-              php vendor/bin/phpunit -v --testsuite Unit,Definition,Community
+              php vendor/bin/phpunit
             fi
 
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -75,18 +75,24 @@ jobs:
       # with the keys in the env variables
       # If the PR was open from a fork (community PR)
       # we get API keys from the API key dealer https://alg.li/api-key-dealer
+      # WE only run the tests on dictionaries on one case to avoid concurrency problems
```
DevinCodes, line 78:
```suggestion
      # we only run the tests on dictionaries on one case to avoid concurrency problems
```

.circleci/config.yml
```diff
@@ -75,18 +75,24 @@ jobs:
...
+                 php vendor/bin/phpunit
+              else
+                 php vendor/bin/phpunit --exclude-group=dictionaries
+              fi
             else
               export CI_PROJ_USERNAME=$CIRCLE_PR_USERNAME
               export CI_PROJ_REPONAME=$CIRCLE_PR_REPONAME
 
-              php vendor/bin/phpunit -v --testsuite Unit,Definition,Community
+              php vendor/bin/phpunit --exclude-group=dictionaries
```
DevinCodes, line 95: Shouldn't we have the same check here? (`"<<parameters.version>>" == "latest" && ...` etc.)

DevinCodes (changes requested):

DevinCodes (approved):

PR #675, opened 2021-04-20
Remove invalid method call

greg0ire:
| Q                 | A
| Bug fix?          | yes
| New feature?      | no    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## Describe your change

I'm removing a call to existent method `ConnectException::hasResponse()`
When a connection fails, there cannot be a response: the client would
have to be connected to get one. Therefore, it would make no sense to
have a `hasResponse()` method on `ConnectException`.

This fixes a crash that can happen in case of connection issues.

### DIFF
--- a/src/Http/GuzzleHttpClient.php
+++ b/src/Http/GuzzleHttpClient.php
@@ -39,17 +39,13 @@ final class GuzzleHttpClient implements HttpClientInterface
                 );
             }
         } catch (ConnectException $e) {
-            if ($e->hasResponse()) {
-                return $e->getResponse();
-            } else {
-                return new Response(
-                    0,
-                    [],
-                    null,
-                    '1.1',
-                    $e->getMessage()
-                );
-            }
+            return new Response(
+                0,
+                [],
+                null,
+                '1.1',
+                $e->getMessage()
+            );
         }
 
         return $response;


greg0ire:
The build failures seem unrelated (429 too many attempts?)

relaypilot:
I am also randomly receiving the following error:

    Call to undefined method GuzzleHttp\\Exception\\ConnectException::hasResponse() at /home/my-website/vendor/algolia/algoliasearch-client-php/src/Http/GuzzleHttpClient.php:`


Using the following versions on PHP 7.4

```
"algolia/algoliasearch-client-php": "^3.0",
"algolia/scout-extended": "^1.17",
"laravel/framework": "^8.12",
"guzzlehttp/guzzle": "^7.0.1",
```

greg0ire:
@chloelbn please review

chloelbn:
I've opened another PR to workaround the CI failing; we'll fix this another way and we'll be able to move forward! 
https://github.com/algolia/algoliasearch-client-php/pull/678

greg0ire:
@chloelbn I think closing and reopening this PR would have been enough: In general, CI usually runs against a merge of the target and the source branch (not sure if that's the case with Circle CI though). So in theory, when reopening, a new merge would have been performed, with the new commit on master that contains your fix :slightly_smiling_face:

PR #682, opened 2021-07-13
chore: rename recommendation client

DevinCodes:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     
| Need Doc update   | yes

## Describe your change
As per our internal specification, this renames the `RecommendationClient` to `PersonalizationClient`. For backwards compatibility you'll still be able to use the `RecommendationClient` until the next major version.

### DIFF
--- a/tests/Integration/PersonalizationClientTest.php
+++ b/tests/Integration/PersonalizationClientTest.php
@@ -6,7 +6,7 @@ use Algolia\AlgoliaSearch\PersonalizationClient;
 
 class PersonalizationClientTest extends BaseTest
 {
-    public function testRecommendationClient()
+    public function testPersonalizationClient()
     {
         $personalizationClient = PersonalizationClient::create(
             getenv('ALGOLIA_APPLICATION_ID_1'),
--- a/tests/Integration/PersonalizationClientTest.php
+++ b/tests/Integration/PersonalizationClientTest.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Tests\Integration;
+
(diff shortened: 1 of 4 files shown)


tests/Integration/PersonalizationClientTest.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Tests\Integration;
+
+use Algolia\AlgoliaSearch\PersonalizationClient;
+
+class PersonalizationClientTest extends BaseTest
+{
+    public function testRecommendationClient()
```
damcou, line 9: I guess you want to rename this as well ? :)

tests/Integration/PersonalizationClientTest.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Tests\Integration;
+
+use Algolia\AlgoliaSearch\PersonalizationClient;
+
+class PersonalizationClientTest extends BaseTest
```
damcou, line 7: Do we want to keep the RecommendationClientTest class ?

damcou (commented):
LGTM, just 2 small comments 💯

tests/Integration/PersonalizationClientTest.php
DevinCodes: I was thinking to keep it, yes. As long as the class is there, we should test it IMO 🙂

damcou: 👍

damcou (approved):

PR #683, opened 2021-07-19
feat: bootstrap new testing strategy

DevinCodes:
| Q                 | A
| Bug fix?          | no
| New feature?      | no    
| BC breaks?        | no     

## Describe your change
This bootstraps a new way to test our API client, where (for PRs) we test the formed requests instead of calling Algolia directly.

### DIFF
--- a/tests/Unit/SearchClientTest.php
+++ b/tests/Unit/SearchClientTest.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Algolia\AlgoliaSearch\Tests\Unit;
+
+use Symfony\Component\Yaml\Yaml;
+
+class SearchClientTest extends AbstractMockClientTest
+{
+    public function testSaveDictionaryEntries()
+    {
+        $client = $this->getClient();
+        $client->saveDictionaryEntries('stopwords', [['objectID' => 'abcdef', 'language' => 'en', 'word' => 'down']]);
+
+        $this->assertRequests($this->loadRequestsForMethod('saveDictionaryEntries'));
+    }
+
+    public function testSearchDictionaryEntries()
+    {
+        $client = $this->getClient();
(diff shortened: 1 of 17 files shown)


VladislavFitz:
Looks really good 👍 
I didn't get how the `loadRequestsForMethod` method works. Where can I find its implementation?

tests/Unit/SearchClientTest.php
```diff
@@ -0,0 +1,87 @@
...
+        );
+    }
+
+    private function loadRequestsForMethod(string $method): array
+    {
+        $request = Yaml::parseFile("./requests_spec/SearchClient/${method}.yml");
+
+        return $request['requests'];
+    }
+}
```
TomKlotzPro, line 87: @VladislavFitz you provide as a String the method name and it will return the content from the yaml you wrote.

However you return just the "request" part of the yaml.

I like this, it's really elegant, we just have to wrote all the different expected request and parse them.

TomKlotzPro:
It's awesome! Really nice work !

Just asking about the YAML request files : 
- Should we all have the YAML files in our repository or could we have a centralised part in the CTS were we just took the files?

tests/Unit/SearchClientTest.php
VladislavFitz: Thank you, I missed it with my cmd + f 😄 Very clear.

DevinCodes:
> Should we all have the YAML files in our repository or could we have a centralised part in the CTS were we just took the files?

@TomKlotzPro I was thinking to have this in a centralized repository, and automatically open PRs on all the clients when we push changes to the production branch. Like that the specs are shared, and we don't rely on git submodules to have it shared.

damcou:
Really nice ! 🚀

PR #696, opened 2021-12-07
fix: psr log

damcou:
| Q                 | A
| Bug fix?          | yes
| Related Issue     | Fix #692

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "psr/http-message": "^1.0",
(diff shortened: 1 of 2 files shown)


greg0ire:
This could be fixed by using a trait in the `DebugLogger`, similar to this: https://github.com/doctrine/dbal/blob/2.13.x/lib/Doctrine/DBAL/Driver/PDOQueryImplementation.php

greg0ire:
Actually since, the class is final, you may just add a `:void` return type declaration, it will not be a BC-break, and should make the signature compatible with both versions.

DevinCodes (approved):

PR #711, opened 2022-10-05
Include input string in message about invalid json

greg0ire:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Related Issue     | n/a
| Need Doc update   | no

## Describe your change

In this PR, we include the input string that is fed to `json_decode`, because it will help understanding issues with it.

This helper is used in a single place, to decode the response coming from Algolia. If the response is invalid, the contents of the input string is lost, making it impossible to troubleshoot the error.

Here is an example of what I am currently experiencing in my application:

`json_decode error: Control character error, possibly incorrectly encoded`

I don't think there is a way for me to troubleshoot this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased](https://github.com/algolia/algoliasearch-client-php/compare/3.3.2...master)
 
+### Added
+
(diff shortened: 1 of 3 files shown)


DevinCodes (approved):
Thank you for this!

DevinCodes:
@damcou do you think we can move forward with this?

damcou (approved):
LGTM

PR #713, opened 2023-01-24
Add referer header.

phuang07:
| Q                 | A
| Bug fix?          | yes
| New feature?      | yes   
| BC breaks?        | yes   
| Related Issue     | Fixed #...  
| Need Doc update   | no

## Describe your change

Add "referer" header before sending out request.

In order to prevent Algolia indexing from dev instances, we need a way to restrict indexing from only the live site.
Algolia provides a feature for API restriction, which can be used to restrict a site to interact with the index by defining a narrow privilege API. 

However, this feature requires to set the referer header in order to work, the site has to send the "referer" header along with the request. 

Currently, this module doesn't set the referer header. This patch fixes this by adding adding header, the referer, before sending out the request.

### DIFF
--- a/src/RequestOptions/RequestOptionsFactory.php
+++ b/src/RequestOptions/RequestOptionsFactory.php
@@ -70,6 +70,9 @@ final class RequestOptionsFactory
                 'X-Algolia-API-Key' => $this->config->getApiKey(),
                 'User-Agent' => UserAgent::get(),
                 'Content-Type' => 'application/json',
+                // For Algolia API restriction, add referer.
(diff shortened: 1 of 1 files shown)


DevinCodes:
Hey @phuang07 ,

Thank you for this PR. We're going to discuss internally if this is something we want to add by default in the API client, as it would require a change in other API clients as well and there are some edge cases to think about.

For the time being, you can work around this in your own code by [setting a custom header](https://www.algolia.com/doc/api-reference/api-methods/set-extra-header/?client=php) with the API client when setting it up. This would allow you to add the referrer by default 🙂

PR #717, opened 2023-07-07
Add support for psr/http-message ^2.0

Stoux:
| Q                 | A
| Bug fix?          | no
| New feature?      | yes    
| BC breaks?        | no     
| Related Issue     | ...  
| Need Doc update   | no

## Describe your change

This introcues support for psr/http-message 2.0 while keeping support for 1.0.

The only change in psr/http-message:2.0 introduces return types. As we support php:^7.2 we can also introduce these to support both 1.0 & 2.0. This also realigns paramater names with their interface definitions to better support named arguments (introduced in php:^8.0).

Allows inclusion in projects that require psr/http-message:^2.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         "ext-curl": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
-        "psr/http-message": "^1.0",
+        "psr/http-message": "^1.0 || ^2.0",
         "psr/log": "^1.0 || ^2.0 || ^3.0",
         "psr/simple-cache": "^1.0 || ^2.0 || ^3.0"
     },
--- a/src/Http/Psr7/BufferStream.php
+++ b/src/Http/Psr7/BufferStream.php
@@ -32,12 +32,12 @@ class BufferStream implements StreamInterface
         $this->hwm = $hwm;
     }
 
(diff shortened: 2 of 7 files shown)


Stoux:
Guzzle 6 tests seem fail because of the lock that is generated, which resolved http-message to the latest (2.0), while guzzle 6 requires http-message (1.0). Works if you just run the require command without generating a .lock first.

Not sure about the code styling errors, as I've only modified existing lines. Was the code styling of those files already incorrect?

tobimori:
Happy to see this merged as I'm currently running into this issue too! @DevinCodes

Stoux:
I've fixed the code style errors. The only remaining is one outside the scope of this PR, otherwise they would have succeeded.

I've also added a commit that allows the Circle CI test to modify the locked dependencies (caused by the initial composer install) when installing Guzzle, this fixes the issue when Guzzle (6) requires http-message 1.0 while 2.0 is installed. Let me know if you want to revert this change.

damcou:
Hello there;

Just a small heads up to confirm that this PR has been spotted and will be reviewed ASAP, probably after we investigate and fix the CI issues in August.
Thanks a lot for submitting this.

PR #719, opened 2023-08-09
Add support for psr/http-message ^2.0

damcou:
This includes: 
- Changes coming from @Stoux (<3) : https://github.com/algolia/algoliasearch-client-php/pull/717
- Fixed linter
- Fixed CI
- Dropped PHP 7.2

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,8 +135,8 @@ workflows:
           version: "7.4"
           http_client: guzzlehttp/guzzle:"^6.0"
       - test:
-          name: 'Guzzle 6 - PHP 7.2'
-          version: "7.2"
+          name: 'Guzzle 6 - PHP 7.3'
+          version: "7.3"
           http_client: guzzlehttp/guzzle:"^6.0"
       - test:
           name: 'Legacy client - PHP 8.1'
@@ -151,6 +151,6 @@ workflows:
           version: "7.4"
           http_client: legacy
       - test:
-          name: 'Legacy client - PHP 7.2'
-          version: "7.2"
+          name: 'Legacy client - PHP 7.3'
+          version: "7.3"
           http_client: legacy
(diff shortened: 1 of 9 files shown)


jacobclarke92:
Just FYI this actually breaks support with psr/http-message ^1  
Probably worth changing composer file to reflect that.  
Solution for me was to pin to 3.3.2  

Reason why it breaks is because of a bunch of bunch of return type incompatibilities e.g.
```
Fatal error:  
Declaration of Algolia\AlgoliaSearch\Http\Psr7\Uri::withScheme(string $scheme): Psr\Http\Message\UriInterface must be compatible with Psr\Http\Message\UriInterface::withScheme($scheme)  
in /Users/jacobclarke/Projects/hutchies/vendor/algolia/algoliasearch-client-php/src/Http/Psr7/Uri.php  
on line 439
```

Stoux:
I've [missed that](https://www.php-fig.org/psr/psr-7/meta/#72-type-additions), didn't test it with the lowest version. It does indeed break on `psr/http-message:~1.0.0` as it sees the new types as more restrictive of the types and thus deem it incompatible.

@damcou Bumping the requirement to `"psr/http-message": "^1.1 || ^2.0",` should suffice, otherwise you would have to revert the 2.0 support.

@jacobclarke92 Upgrading `psr/http-message` to 1.1 (or 2.0) should fix it, if you don't have any dependencies that lock it to 1.0 of course.

damcou:
> I've [missed that](https://www.php-fig.org/psr/psr-7/meta/#72-type-additions), didn't test it with the lowest version. It does indeed break on `psr/http-message:~1.0.0` as it sees the new types as more restrictive and thus deem it incompatible.
> 
> @damcou Bumping the requirement to `"psr/http-message": "^1.1 || ^2.0",` should suffice, otherwise you would have to revert the 2.0 support.

Just released v3.4.1 with this changes which should indeed solve the issue.
Thanks for the suggestion !

PR #745, opened 2025-01-13
Address PHP 8.4-specific deprecation

greg0ire:
Since PHP 8.4, implicitly marking a parameter as nullable is deprecated.


Fluf22 (approved):
Thank you!

